PR #4, opened 2015-03-16
Docs for the deployment process

nibalizer:


docs.html
```diff
@@ -66,8 +66,14 @@
 <h2>
 <a name="publishing" class="anchor" href="#publishing"><span class="octicon octicon-link"></span></a>Publishing a module</h2>
 
-<p>Push a signed tag to the module then ping an admin (@daenney, @igalic, @nibalizer) and they will publish the module.</p>
-<p>TODO: document module publish process for admins</p>
+<p>Forge publishing is handled by travis and puppet-blacksmith.</p>
+<p>To enable a module to be pushed, cargo-cult the 'deploy' block from puppetcommunity-extlib travis.yml.</p>
```
igalic: i'd rather we setup a modulesync here, and, asap, too ;)

docs.html
```diff
@@ -66,8 +66,14 @@
...
+<p>The 'secure' line in the travis.yml is created by:</p>
+<pre>travis encrypt -r puppet-community/puppetcommunity-extlib $the_forge_password</pre>
+<p>Obivously we'll need to redo this whenever we roll the password</p>
```
igalic: nothing is obvious.

docs.html
```diff
@@ -66,8 +66,14 @@
 <h2>
 <a name="publishing" class="anchor" href="#publishing"><span class="octicon octicon-link"></span></a>Publishing a module</h2>
 
-<p>Push a signed tag to the module then ping an admin (@daenney, @igalic, @nibalizer) and they will publish the module.</p>
-<p>TODO: document module publish process for admins</p>
+<p>Forge publishing is handled by travis and puppet-blacksmith.</p>
+<p>To enable a module to be pushed, cargo-cult the 'deploy' block from puppetcommunity-extlib travis.yml.</p>
+<p>The 'secure' line in the travis.yml is created by:</p>
+<pre>travis encrypt -r puppet-community/puppetcommunity-extlib $the_forge_password</pre>
```
igalic: where do people get the password from?

<insert link to gpg secured password>
<add info that, if you're not on the list of people entrusted with the password, ask one of them>

docs.html
```diff
@@ -66,8 +66,14 @@
...
+<p>Obivously we'll need to redo this whenever we roll the password</p>
+
+<br>
+<p>Gem publishing is handled similarly, except there is not a unified user.</p>
```
igalic: should there be? does that make sense? does it make things simpler, or more prone to attacks / single points of failure?

docs.html
mkrakowitzer: I would replace the word cargo-cult, I had to look it up on wikipedia, it's a little derogatory, perhaps just use 'copy' instead.

nibalizer:
Fixed

igalic:
*merged*

PR #6, opened 2015-03-21
Actually link to the docs page

nibalizer:
I'm not sure why we didn't have this before


igalic:
my thinking was to finish it, and then pull the docs into index.html, cuz, it's not that much…

nibalizer:
I don't really care that much, but I do see docs expanding constantly. The index page being a concise landing page is good imho.

igalic:
ack

PR #32, opened 2016-03-01
Add link to modulesync_config directory.

nanliu:


daenney:
Can we change the "the voxpupuli module sync configuration" to simply say "our module sync configuration"?

nanliu:
Updated per request.

PR #41, opened 2016-06-09
Clarify version numbering

alexjfisher:
Only the minor version should be bumped if the module is pre version 1.0 and ruby 1.8 support has been dropped.


igalic:
imo, extlib should be bumped to 1.0.0
in general, i'm all for making modules we adopt 1.0.0 if they aren't yet.

PR #42, opened 2016-06-10
add transcript of my puppetcamp paris&london 2016 talk

igalic:


igalic:
from what i gather, i'll have to fix-up the images to… not be GINORMOUS.

rnelson0:
GH doesn't show file sizes, at least that I can see. The build did fail, but I'm not certain all of those failures are legitimate.

daenney:
The error to fix is this one:

```
             Error: could not read file /home/travis/build/voxpupuli/voxpupuli.github.io/_posts/2016-06-10-voxpupuli-opensource-speedtrain.md: (<unknown>): mapping values are not allowed in this context at line 3 column 17
```

```
title: Voxpupuli: Opensource Speedtrain
```

This is YAML so that `:` causes the parser to think you're defining a Hash. Just escape it with a `\` or put quotes around the whole string.

daenney:
The reason for it failing on the images is because the URL that is getting built up is `http://localhost:4000/blog/2016/06/10/voxpupuli-opensource-speedtrain/static/images/8bit-vox.png` which obviously doesn't exist.

This happens because you're using `8bit vox` where it should be `8bit vox`

daenney:
Great post by the way, totally ❤️ it.

daenney:
If you rebase now that #44 is merged you should also no longer get errors from the html proofer when adding a new post that the external link for it isn't live yet.

daenney:
And with #46 merged we now no longer check for voxpuupli.org URL's. So fix the two issues I mentioned and rebase on master should cause this to go green.

PR #45, opened 2016-06-11
Update a few components of the website

daenney:
* Add a favicon
* Update theme and Font Awesome
* Point Ci at Travis, VPCI appears mostly broken


rnelson0:
`*  External link https://voxpupuli.org/favicon.ico failed: 404 No error`
Didn't #44 fix this?

daenney:
No, #44 only fixes it for posts and nav links, not any random asset. I'm working on that one next.

rnelson0:
OK, that seems like the only non-problem, so here we go!

PR #60, opened 2016-08-12
"Use javascript to dynamically get stats for the main page"

nibalizer:


daenney:
Since the only difference between `index.html` and `default.html` is the inclusion of this one extra, tiny piece of javascript I'd rather we just include that JS everywhere, it's one less template to maintain.

We could also conditionally include it if we detect that we're on the front page.

nibalizer:
I'd rather not run the js every time (makes the client load a request from a remote server) but the conditional stuff you mention is great. Will incorporate.

PR #62, opened 2016-08-25
add top_scope_facts and legacy_facts linters

mmckinst:
useful to take advantage of the structured facts and `$facts` hash in
puppet 4 but puppet 3 users can also make use of them if they've
configured puppet correctly or are running a new version of facter

* top_scope_facts - lints for facts like `$::operatingsystem` and
replaces them with `$facts['operatingsystem']` .

* legacy_facts - lints for so called legacy facts like
`$facts['operatingsystemmajrelease']` or `$::operatingsystemmajrelease`
and replaces them with the new fancy structured fact like
`$facts['os']['release']['major']`

ref https://github.com/voxpupuli/plumbing/issues/21

### DIFF
--- a/_data/tools/puppet-lint.json
+++ b/_data/tools/puppet-lint.json
@@ -145,6 +145,16 @@
       "description": "A puppet-lint plugin to check if resources exist outside of a class or defined type."
     },
     {
+      "name": "top_scope_facts ",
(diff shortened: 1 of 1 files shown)


daenney:
Thanks @mmckinst!

PR #65, opened 2016-08-28
"Don't specify # of collaborators or repos"

nibalizer:


roidelapluie:
:-1: we should update the numbers for people with noscript

nibalizer:
We made it dynamic so that we wouldn't have to update those numbers through pull requests frequently(because we know we are lazy and we won't do it). I think putting nothing there is a good way to tell our friends with noscript to whitelist our 'get the # of repos' script.

roidelapluie:
We could remove that. It is not relevant for membres as your only get 30 with thé API...

daenney:
I think it's actually nice to show if we can fetch it. So what I would suggest instead is that we remove the HTML completely, so that people with JS don't even see the placeholder divs, and then use JS fully to insert that.

roidelapluie:
I think we can fetch the data at the travis level. When we push, the number gets updated.

daenney:
Probably, but that still relies on us doing relatively regular updates to the site. Looking at the commit cadence there's at least a change every month but there's no guarantee that that'll continue happening.

Though I'll grant you it's much nicer than having it completely static. I would still keep the JS though since that gives people with a JS-enabled browser the most up to date numbers.

ffrank:
If there are no objections, I'm going to close this in favor of a feature request to have the static numbers updated semi-automatically, as @roidelapluie suggested.

PR #77, opened 2016-12-05
Introduce Security

roidelapluie:
Signed-off-by: Julien Pivotto <roidelapluie@inuits.eu>


bastelfreak:
:+1: from my side. @rnelson0 is there any specific reason for the label?

rnelson0:
@bastelfreak that's from the waffle.io board. I can disable it on this board when I get back on Friday if you'd like.

PR #78, opened 2016-12-05
Change from Best to Good practices ; link to gpg signing.

roidelapluie:
Signed-off-by: Julien Pivotto <roidelapluie@inuits.eu>


bastelfreak:
thanks @roidelapluie!

PR #79, opened 2016-12-22
Blog post: putting down puppet 3

nibalizer:


_posts/2016-12-22-putting-down-puppet-3.md
```diff
@@ -0,0 +1,19 @@
...
+layout: post
+title: Putting down Puppet 3
+date: 2016-12-22
+github_username: nibalizer
+twitter_username: nibalizer
+---
+
+Puppet 3 is officially end of life on Jan 1, 2017. That is 8 days from today. Vox Pupuli's plan for deprecating Puppet 3 support in modules is as follows:
+
+* As of Jan 1, 2017, any module can accept Puppet 4 code on the master branch.
```
bbriggs: s/any module can accept/we will gradually begin deploying/

bbriggs: I think that provides a clearer statement that we are officially and actively deprecating 3 in favor of 4

nibalizer: I don't agree with that change. 'Gradually' is a weird word that implies some kind of slow change, I'd rather not confuse people with 'are we being gradual?' I also don't like 'deploying' because we are not deploying anything.

I used 'any module can accept' because that is something our collaborators can look at as documentation and say 'yes I can merge this puppet 4 only code because the blog post says so'

bbriggs: Gotcha. I'm fine with that then.

PR #91, opened 2017-01-18
add github PR stats jekyll plugin

foxxx0:
Also introduce the "stats" config key for backup values
This way only the _config.yml needs to be modified when manually
updating the values.

The only thing left to do is:
adding the encrypted GH_CLIENT_ID and GH_CLIENT_SECRET ENV vars to .travis.yml.

Maybe @bastelfreak can help out here? :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,14 @@
 language: ruby
 sudo: false
 before_install:
-- gem install bundler
+  - gem install bundler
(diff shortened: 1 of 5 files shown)


bastelfreak:
closing this since we merged it in https://github.com/voxpupuli/voxpupuli.github.io/pull/94

PR #96, opened 2017-01-18
[WIP] Let travis update our start page by committing to github

bastelfreak:


.gitignore
```diff
@@ -3,5 +3,7 @@ _site/
 .sass-cache/
 .jekyll-metadata
 Gemfile.lock
+_config.yml.bak
 .vendor/
 .bundle/
+id_ed25519
```
igalic: what is this (for)?

tasks/support/github_pr_stats.rb
```diff
@@ -0,0 +1,116 @@
...
+    req = Net::HTTP::Get.new(url.to_s)
+    req['Accept'] = 'application/vnd.github.v3+json'
+    req['User-Agent'] = 'voxpupuli.org github API'
+    res = Net::HTTP.start(url.host, url.port, use_ssl: true) { |http| http.request(req) }
+
+    raise GithubAPIError.new, "API response status: #{res.code}" if res.code.to_s =~ %r{(4|5)\d\d}
+
+    # see if there are more pages
+    next_page = nil
+    if res['link']
```
igalic:
```ruby
return [next_page, res.body] unless res['link']
```
to remove on layer of indentation below

tasks/support/github_pr_stats.rb
```diff
@@ -0,0 +1,116 @@
...
+      res['link'].split(',').each do |link|
+        # also catch possibly nested quoting
+        if link =~ %r{rel=(\\|)"next(\\|)"}
+          next_url = URI.extract(link)[0]
+          next_page = next_url.match(%r{(&|\?)page=(?<page>[0-9]+)})[:page]
```
igalic: could it be that the github API is absolutely atrocious?

tasks/support/github_pr_stats.rb
```diff
@@ -0,0 +1,116 @@
...
+      log("  [+] processing #{repo['name']}...")
+      pull_url = repo['pulls_url'].chomp('{/number}')
+      pull_url += '?state=open&per_page=100'
+      pull_url += @auth_suffix
+      next_page = 1
+      loop do
+        next_page, pull_requests = query(url: pull_url, page: next_page)
+        @pull_reqs += pull_requests.length if pull_requests.is_a?(Array)
+        break unless next_page
+      end
```
igalic: why the `loop`/`break`, when you could just `while ! next_page.nil?` instead?

tasks/support/github_pr_stats.rb
```diff
@@ -0,0 +1,116 @@
...
+    return nil unless @auth_suffix
+
+    log('fetching data from github API, this may take a while...')
+
+    begin
+      next_page = 1
+      loop do
+        next_page, repositories = query(url: @api_url, page: next_page)
+        process_repos(repositories)
+        break unless next_page
```
igalic: same as https://github.com/voxpupuli/voxpupuli.github.io/pull/96/files#r97201010

tasks/support/github_pr_stats.rb
```diff
@@ -0,0 +1,116 @@
...
+      FileUtils.copy_file(@config_file, @config_file + '.bak')
+
+      log("updating #{@config_file}")
+      @config['stats']['prs'] = @pull_reqs
+      File.open(@config_file, 'w') { |f| f.write(@config.to_yaml) }
+      log('done')
+      return @pull_reqs
+    rescue GithubAPIError => err
+      p(err)
+      return nil
```
igalic: this feels thoroughly dirty…
if you're packing all this code into a single begin block, you could just as well have the query function do error logging & exit… but then it should have to be a `query!` function which might seem… unintuitive ;)

tasks/update_pr_stats.rb
```diff
@@ -0,0 +1,78 @@
...
+    gh_pr_stats = GithubPRStats.new
+    result = gh_pr_stats.update
+
+    next unless ENV['CI'] && ENV['CI']
+    next unless ENV['TRAVIS'] && ENV['TRAVIS']
+    next if result.nil?
+
+    identity_file = './id_ed25519'
+    unless File.exist?(identity_file)
+      puts('deploy key not found, skipping git commit & push...')
```
igalic: this should use logging infrastructure — maybe this should even be an error

tasks/update_pr_stats.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    # permissions
+    system("chmod 0700 #{Dir.pwd}")
+    system("chmod 0700 #{Dir.home}")
+    system("chmod 0600 #{identity_file}")
+    system('umask 077')
+
+    # create symlink for the git ssh wrapper
+    system("ln -sf $(readlink -f #{identity_file}) #{Dir.home}/gh_deploy_key")
+    system("chmod 0600 #{Dir.home}/gh_deploy_key")
```
igalic: why not use `FileUtils` here?

tasks/update_pr_stats.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    git_diff = `git diff --stat _config.yml`
```
igalic: should we use something "better" than raw git commands?

_config.yml
```diff
@@ -1,42 +1,42 @@
...
+- README.md
+- CNAME
+- README.md
+- vendor
+- tasks
 stats:
   collaborators: 30
   repos: 119
-  pr: 183
+  prs: 167
```
bastelfreak: the original values were hardcoded a long long time ago. This PR allows us to fetch this data from the github API by looping over each of our repos and asking for the amount of open PRs.

_config.yml
```diff
@@ -1,42 +1,42 @@
...
+- Gemfile
+- Gemfile.lock
+- LICENSE
+- README.md
+- CNAME
+- README.md
+- vendor
+- tasks
 stats:
   collaborators: 30
```
bastelfreak: the API only lists users with public membership, most of our users have a private membership.

_config.yml
igalic: so they're just placeholders then?

bastelfreak: jekyll parses them to embed them into the html code which is displayed on our front page.

tasks/support/github_pr_stats.rb
foxxx0: yes, but they tell you everywhere to use that HTTP Link header to traverse the result pages...

.gitignore
foxxx0: this is the ssh identity file (deploy key) which allows travis to push to this repository, to prevent any potential problems we just exclude it from being tracked by git

_config.yml
foxxx0: more like default values.

jekyll will use these values to render the html pages, so that even if travis was unable to update the values, an old version will be rendered.

tasks/support/github_pr_stats.rb
foxxx0: you would need to specify the query() call twice, once before the loop and once within, this looks cleaner to me imho.

tasks/update_pr_stats.rb
foxxx0: if there are PRs from unprivileged users (i.e. users without write access to this repo) travis won't expose any encrypted information (i.e. env vars and files) into the build.

this prevents sensitive information from being leaked.

but if you want to allow for clean/successful builds of unprivileged PRs, you need to skip certain parts.

tasks/update_pr_stats.rb
foxxx0: open for suggestions, wanted to keep dependencies at a minimum.

tasks/support/github_pr_stats.rb
foxxx0: open for suggestions :)

tasks/support/github_pr_stats.rb
foxxx0: sounds good

tasks/update_pr_stats.rb
igalic: i'd rather keep reinventing wheels to a minimum :D

roidelapluie:
Please note that now travis can run jobs daily .. so we could have up to date facts

roidelapluie:
https://docs.travis-ci.com/user/cron-jobs/

bastelfreak:
@foxxx0 let's just close this

PR #103, opened 2017-02-10
bump modulesync version from 0.16.10 -> 0.16.11

bastelfreak:


dhoppe:
The file `.travis.yml` has duplicate entries.

bastelfreak:
ah what the hell, why does it fail again...

dhoppe:
Line 4 at `.travis.yml` needs to be removed?

bastelfreak:
@dhoppe this looks really strange. I can't reproduce the errors locally.

.travis.yml
```diff
@@ -1,11 +1,13 @@
...
 sudo: false
+cache: bundler
 before_install:
-  - gem install bundler
   - gem update --system
+  - gem install bundler
 dist: trusty
 rvm:
-  - 2.4.0
+  - 2.3.0
```
bastelfreak: the json gem is in version 2.X in the ruby2.4 version, but one of our dependencies needs 1.X :(
pinning json didn't work out, so I had to downgrade our ruby version

PR #123, opened 2018-02-07
Migration doc clarity

rnelson0:
You add people, not modules, to teams.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -49,7 +49,7 @@ You will have someone by your side in this process. The general flow is to…
 * Release a copy of your module to the 'puppet' forge account.
 * Add the module to our [modulesync setup](https://github.com/voxpupuli/modulesync_config/blob/master/managed_modules.yml)
 * Add the module to our [plumbing repository](https://github.com/voxpupuli/plumbing/blob/master/share/modules)(handles travis secrets)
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -49,7 +49,7 @@ You will have someone by your side in this process. The general flow is to…
 * Release a copy of your module to the 'puppet' forge account.
 * Add the module to our [modulesync setup](https://github.com/voxpupuli/modulesync_config/blob/master/managed_modules.yml)
 * Add the module to our [plumbing repository](https://github.com/voxpupuli/plumbing/blob/master/share/modules)(handles travis secrets)
-* Ask one of the admins to add the module to the collaborators Team on github.
+* Ask one of the admins to add the previous module owner(s) to the collaborators Team on github.
```
bastelfreak: those are two different steps. We shouldnt replace it. The old step is always needed, the new one is only needed if they want to have merge access.

rnelson0: Should this instead say to add the collaborators team to the module’s teams? You cannot add modules to teams, so the existingnlanguage is confusing.

bastelfreak: You can add a team to a module, in the module settings, or add a module to a team: https://github.com/orgs/voxpupuli/teams/collaborators/repositories

PR #125, opened 2018-03-12
Update our Review Guidelines

bastelfreak:


docs/index.md
```diff
@@ -115,12 +115,17 @@ Travis will then kick off a build against the new tag created and deploy that bu
...
-* Correct email address used in the commits (if so, travis displays the avatar next to the commit)?
+* Correct email address used in the commits (if so, github displays the avatar next to the commit)?
 * Is this a bugfix, modulesync, breaking change, enhancement, docs update? Label it with `bug`, `modulesync`, `backwards-incompatible`, `enhancement`, `docs`
 * Are updates to the README.md needed but missing? Label it with `needs-docs`
+* Has the file documented params or examples in the header? This needs to be updated as well
 * Are there merge conflicts? Add the `needs-rebase` label
 * Does it need additional tests? Add the `needs-tests` label
-* Are new parameters introduced? They should have datatypes
+* Does it have failing tests? Add the `tests-fail` label
+* Are new parameters introduced? They must datatypes
```
ekohl: s/must/must have/

docs/index.md
```diff
@@ -115,12 +115,17 @@ Travis will then kick off a build against the new tag created and deploy that bu
...
 * Are facts used? They should only be accessed via `$facts[]` or [fact()](https://github.com/puppetlabs/puppetlabs-stdlib#fact) from stdlib, but not topscope variables
+* Varibles shouldn't be accessed via topscope: $::$modulename::$param. Instead do: $modulename::$param
```
ekohl: s/$::$/$::/

docs/index.md
```diff
@@ -115,12 +115,17 @@ Travis will then kick off a build against the new tag created and deploy that bu
...
 * Are datatypes from stdlib used? Ensure that lowest supported stdlib version is 4.13.1. Check if a newer version introduced the used datatype
-* Are hiera yaml files added for /docs? Ensure that the lowest supported Puppet version is 4.10.9
+* Are hiera yaml files added for data-in-modules? Ensure that the lowest supported Puppet version is 4.10.9
+* Are there new params with datatype Hash or Array? If possible, they should default to empty Hash/Array instead of undef
```
ekohl: I also like stricter data types. ```Array[String]``` over ```Array``` if possible.

docs/index.md
```diff
@@ -115,12 +115,17 @@ Travis will then kick off a build against the new tag created and deploy that bu
...
+* Are there new params with datatype Boolean? If possible, they should default to True/False instead of undef
```
ekohl: I'd say ```Optional[Boolean]``` but there are valid cases. A config key not showing up in generated config can be valid.

docs/index.md
```diff
@@ -115,12 +115,17 @@ Travis will then kick off a build against the new tag created and deploy that bu
 
 There are a few things that can be checked if you review a pull request against one of our modules:
 
-* Correct email address used in the commits (if so, travis displays the avatar next to the commit)?
+* Correct email address used in the commits (if so, github displays the avatar next to the commit)?
 * Is this a bugfix, modulesync, breaking change, enhancement, docs update? Label it with `bug`, `modulesync`, `backwards-incompatible`, `enhancement`, `docs`
 * Are updates to the README.md needed but missing? Label it with `needs-docs`
+* Has the file documented params or examples in the header? This needs to be updated as well
```
ekohl: Should we start using https://github.com/voxpupuli/puppet-lint-param-docs? It needs a whitelist of files to check (like https://github.com/theforeman/puppet-foreman/blob/645a48c15dbf30901179afb198d4ab392484335d/Rakefile#L21-L24)

docs/index.md
```diff
@@ -115,12 +115,17 @@ Travis will then kick off a build against the new tag created and deploy that bu
...
 * Are there merge conflicts? Add the `needs-rebase` label
 * Does it need additional tests? Add the `needs-tests` label
-* Are new parameters introduced? They should have datatypes
+* Does it have failing tests? Add the `tests-fail` label
+* Are new parameters introduced? They must datatypes
 * Are facts used? They should only be accessed via `$facts[]` or [fact()](https://github.com/puppetlabs/puppetlabs-stdlib#fact) from stdlib, but not topscope variables
+* Varibles shouldn't be accessed via topscope: $::$modulename::$param. Instead do: $modulename::$param
 * Are datatypes from stdlib used? Ensure that lowest supported stdlib version is 4.13.1. Check if a newer version introduced the used datatype
-* Are hiera yaml files added for /docs? Ensure that the lowest supported Puppet version is 4.10.9
+* Are hiera yaml files added for data-in-modules? Ensure that the lowest supported Puppet version is 4.10.9
```
ekohl: A rake check would be nice btw.

docs/index.md
bastelfreak: fixed

docs/index.md
bastelfreak: yeah the example was a bit confusing

docs/index.md
bastelfreak: fixed

docs/index.md
bastelfreak: correct. I updated the text

docs/index.md
bastelfreak: I would like to enforce puppet-strings docs for new params, but most modules don't have existing docs :(. I'm not sure how helpful the linter plugin would be?

docs/index.md
bastelfreak: yes :(

docs/index.md
EmilienM: s/Varibles/Variables/

bastelfreak: fixed

docs/index.md
```diff
@@ -115,12 +115,17 @@ Travis will then kick off a build against the new tag created and deploy that bu
...
+* Does it have failing tests? Add the `tests-fail` label
+* Are new parameters introduced? They must have datatypes
 * Are facts used? They should only be accessed via `$facts[]` or [fact()](https://github.com/puppetlabs/puppetlabs-stdlib#fact) from stdlib, but not topscope variables
+* Varibales shouldn't be accessed via topscope: $::modulename::$param. Instead do: $modulename::$param
 * Are datatypes from stdlib used? Ensure that lowest supported stdlib version is 4.13.1. Check if a newer version introduced the used datatype
-* Are hiera yaml files added for /docs? Ensure that the lowest supported Puppet version is 4.10.9
+* Are hiera yaml files added for data-in-modules? Ensure that the lowest supported Puppet version is 4.10.9
+* Are there new params with datatype Hash or Array? If possible, they should default to empty Hash/Array instead of undef. You can also enforce the datastructure like Array[String]
+* Are there new params with datatype Boolean? If needed, they should default to True/False instead of undef. This is a tricky decision which needs careful reviewing. Sometimes a default is the better approach, sometimes undef
```
alexjfisher: I think there are too many exceptions.  In many cases defaulting to `undef` and then not for example writing a line in a config file is the better option.

docs/index.md
```diff
@@ -115,12 +115,17 @@ Travis will then kick off a build against the new tag created and deploy that bu
...
 * Is this a bugfix, modulesync, breaking change, enhancement, docs update? Label it with `bug`, `modulesync`, `backwards-incompatible`, `enhancement`, `docs`
 * Are updates to the README.md needed but missing? Label it with `needs-docs`
+* Has the file documented params or examples in the header? This needs to be updated as well
 * Are there merge conflicts? Add the `needs-rebase` label
 * Does it need additional tests? Add the `needs-tests` label
-* Are new parameters introduced? They should have datatypes
+* Does it have failing tests? Add the `tests-fail` label
+* Are new parameters introduced? They must have datatypes
 * Are facts used? They should only be accessed via `$facts[]` or [fact()](https://github.com/puppetlabs/puppetlabs-stdlib#fact) from stdlib, but not topscope variables
+* Varibales shouldn't be accessed via topscope: $::modulename::$param. Instead do: $modulename::$param
```
alexjfisher: spelling of 'variables'.  Anyway, is this true?  I thought it was classes that definitely didn't need to be accessed by topscope anymore.  Variables, not sure.

alexjfisher:
Would it be better to suggest amendments to the official style guide?  Then maybe write puppet-lint plugins to enforce the agreed style.

docs/index.md
ekohl: Should it be ```$::modulename::param``` rather than ```$param```?

bastelfreak:
@alexjfisher on a long term, yes. But that needs time. I think collecting our opinions in a central place is a good start.

docs/index.md
bastelfreak: My phrasing isn't perfect. I will try to rephrase it.

bastelfreak: updated it

docs/index.md
bastelfreak: What I wanted to say is: You shouldn't use $:: anymore, no matter if it's a class, param or fact.

bastelfreak: updated it

alexjfisher: Are you sure?  I thought it was just classes where this was 100% true.  and what about parameters from the ENC?

ekohl: Aren't parameters from the ENC presented as facts? Haven't checked this tbh but I see them in the same way.

rnelson0: There is no dynamic scoping of variables or class names anymore.

PR #128, opened 2018-03-28
add a note about multiple values for a parameter

bastelfreak:


juniorsysadmin:
I don't like this, mainly because I don't like `Variant` and I don't like logic complexity moved away from the user back to the module.

bastelfreak:
@juniorsysadmin what would be your suggestion? If a parameter can have multiple values it should always be a list/array? I'm fine with both, I just think we should decide on one side and write it down.

juniorsysadmin:
> If a parameter can have multiple values it should always be a list/array?

Yeah, that would be my preference.

bastelfreak:
@voxpupuli/collaborators I would like to get a few opinions here :)

jhoblitt:
https://www.youtube.com/watch?v=AKXlhpeb6wI

JacobHenner:
Truly mixed on this, as it would add unnecessary complexity to parameter typing and handling, but would simplify module use for the user. For the sake of consistency and clearly indicating that a parameter accepts one or more values, it's probably best to require an array. A user who looks at an example which uses a singular value (e.g. a string) might otherwise be confused if he/she does not search for the datatype, the parameter's usage, or additional examples/unit tests.

dhollinger:
I lean more towards ease of use for the user rather than simplicity for the developer. If we apply this _only_ to parameters that explicitly take "one or more" values, then I would prefer that.

vStone:
There is not much to do for a developer to deal with this... `[$var].flatten()` and done. Users of our modules are our target audience no? we should make their life easier.

oranenj:
In the case of one or more values, I vote for the variant wrapper approach. It makes the interface more intuitive for the user with negligible code complexity impact.

bastelfreak:
@vStone I like your idea and updated the docs with that pattern.

JacobHenner:
I'm not sure allowing both strings and arrays of strings makes users' lives easier - many users (especially newer ones) will be using examples from a module to guide their use of it. If the example includes a singular value where an array is also acceptable, they might not immediately realize the parameter accepts more than one value.

bastelfreak:
@JacobHenner this needs to be documented properly in the README.md / the puppet-strings docs, no matter which way we choose.

JacobHenner:
@bastelfreak Indeed. Perhaps that should be noted in this guideline?

ekohl:
I'm leaning to prefer just arrays on new parameters, but for compatibility reasons it's fine to keep both in there. I have thought about a `OneOrMore[T]` type but I don't know if the puppet type system allows for this. It'd be great if the Puppet language itself could box a single string into an array of strings.

bastelfreak:
@ekohl I had a quick discussion with @binford2k about this. My idea was that a lambda function instead of a datatype would be cool.  That would easily allow us to build such validations and transformations.

ekohl:
> @ekohl I had a quick discussion with @binford2k about this. My idea was that a lambda function instead of a datatype would be cool. That would easily allow us to build such validations and transformations.

I'd be interested in seeing such a solution.

binford2k:
> There is not much to do for a developer to deal with this... [$var].flatten() and done.

The thing that bothers me about this approach is that it's the old `$real_variable` pattern cropping up again--one of the things that having a robust type system was supposed to help eradicate. I want the type system to have the ability to do something like autobox the parameter into a specified type. Eg, the proposed `Oneormore[String]` should match `Variant[String, Array[String]]` and automatically promote the value to an `Array` if needed. I don't expect much traction on this though; when I requested a similar feature for `Optional`, it was rejected.

reidmv:
@binford2k there may be room for requesting functionality like you're describing in the form of automatic "casting". E.g. for code like this:

```puppet
class majik (
  Sensitive $inkantation,
) {
  # REDACTED
}
```

It's natural to want a supplied value of type String to be automatically type cast to Sensitive.

If that argument holds and gets traction in engineering, it's a natural and simple extension to say you would like a type Oneormore[T] which is basically Array[T], but can automatically cast T to Array[T].

Might get more consideration if it's framed that way instead of something more scary-sounding (to engineering) like lambda functions.

ekohl:
@reidmv I was very surprised this was not already the case. This made the functionality of the `Sensitive` type very limited for me. Within the templating scope it should also automatically unbox back to a string, but not log the change. I realize that this may be hard within ERB (or even EPP), but *that* would make it a very useful type.

docs/index.md
```diff
@@ -159,3 +159,4 @@ There are a few things that can be checked if you review a pull request against
 * Files should always terminate with a newline if possible, with an exception being file or template fragments like those used with concat. This is the [POSIX standard](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206), and some tools don't handle the lack of a terminating newline properly
 * You can merge your own PR if it was approved by someone else and travis is green. Don't merge if either one of those conditions are not true
 * It's okay to approve code regardless if travis is still running or not. The code won't be merged if travis fails after the PR got approved
+* If you can supply one or multiple values for an attribute it's common practice to enforce the datatype for one value and an array of that datatype. An example for string is `Variant[String[1],Array[String[1]]]`. This can be used in the Puppet code as `$var].flatten()`
```
dhollinger: It appears there is a typo at the end of the sentence, should `$var].flatten()` be `[$var].flatten()`?

PR #129, opened 2018-03-29
explain params that map to config options

bastelfreak:
This is the outcome of a discussion about https://github.com/voxpupuli/puppet-nginx/pull/1189 on the IRC.


alexjfisher:
I'm unconvinced about this.  Looking back at the IRC logs, at least a couple of people preferred the use of https://github.com/puppetlabs/puppetlabs-stdlib#bool2str

PR #134, opened 2018-04-26
add a statement about newlines

bastelfreak:


bastelfreak:
I would like to keep this open for a few days to get some feedback. Just in case anybody is against it

bastelfreak:
We need to clarify this. We're talking about a \n at the last character in the last line. Not about an empty line.

PR #137, opened 2018-05-05
add a note about private classes

bastelfreak:


juniorsysadmin:
I have mixed feelings about making this a blanket rule. It should generally be assumed that init.pp is the entry point and all other classes are private.

bastelfreak:
@juniorsysadmin I updated it

PR #157, opened 2018-12-26
add blogpost for puppet4 deprecation

bastelfreak:
@voxpupuli/collaborators I would like to get some feedback here :)

### DIFF
--- a/_posts/2019-01-03-dropping-puppet4.md
+++ b/_posts/2019-01-03-dropping-puppet4.md
@@ -1,0 +1,23 @@
+---
+layout: post
+title: Dropping Puppet 4 support
+date: 2019-01-03
+github_username: bastelfreak
+twitter_username: BastelsBlog
+---
+
+Puppet 4 reached End of Life on 2018-12-31. There isn't an official support
+matrix for the FOSS components. For Puppet enterprise this can be found
+[here](https://puppet.com/misc/puppet-enterprise-lifecycle). PE 2018.1 is the
+oldest currently supported version. It currently ships
+[Puppet 5.5.8](https://puppet.com/docs/pe/2018.1/component_versions_in_recent_pe_releases.html#puppet-enterprise-agent-and-server-components).
+
+Vox Pupuli will remove Puppet 4.x from the metadata.json and from the Travis CI
+testmatrix after 2019-01-01. Our lowest supported Puppet version will be 5.5.8.
+This deprecation will be followed by a major version bump of each Puppet
+module. We won't do releases just for this change, but rather follow our normal
+release schedule.
+
+For the Puppet 3 deprecation we created a branch called `puppet3` with the
+legacy codebase. This created quite some hassle and confusion, so we won't do
+this for Puppet 4.


ekohl:
I still disagree that we should be so aggressive in dropping support. IMHO dropping support is more effort than keeping it. This will just be a huge batch of PRs that only drop Puppet 4 and releases. At this moment I'm unsubscribing from some repositories because of noise. I'm afraid I'll unsubscribe from a lot more.

I'm much more in favor of a lazy approach where we drop it if we need/want features from newer versions.

dhollinger:
While I don't fully agree with @ekohl on not dropping support for puppet 4.x, I do think we need to roll this out more slowly to give projects and companies time to adjust their environments. I'd say the blog post should go up and note that we will have a goal of removing Puppet 4 support from **NEW Releases** by the end of Q1.

How do people feel about that. It allows us time to roll out changes. That said, I think we need to or we will end up in the kind of support hell that Puppetlabs has with some of their modules. Simply put changes between Puppet 4 to 6 will eventually make it impossible to support all 3 major releases in a single module. We also are not RedHat, Canonical, or even Puppetlabs themselves. I think we need to treat our support matrix similar to that of CentOS - when upstream drop support for a release we do too.

At some point, we need to start forcing the world to become more agile. It is actually worse for the larger community as a whole to hold back on dropping support for the legacy organizations not willing to change.

tuxmea:
I don't see the need to push us under pressure to remove Puppet 4 support from all modules under hurry. I like the suggestion from @dhollinger to drop Puppet 4 support for new releases by end of Q1 2019. 
As mentioned in the blog post, I don't like having a branch living around with Puppet 4 support. People who need Puppet 4 can use the latest tag with Puppet 4 support.

_posts/2018-12-27-dropping-puppet4.md
```diff
@@ -0,0 +1,22 @@
...
+
+Puppet 4 will be end of life by the December 31 2018. There isn't an official
+support matrix for the FOSS components. For Puppet enterprise this can be found
+[here](https://puppet.com/misc/puppet-enterprise-lifecycle). PE 2016.4 is the
+oldest currently support version. It currently ships
+[Puppet 4](https://puppet.com/docs/pe/2017.3/component_versions_in_recent_pe_releases.html).
+
+By January 1 2019, Puppet Enterprise 2018.1 will be the oldest supported
+version by Puppet Inc. It contains Puppet 5.5.8. Vox Pupuli will remove Puppet
+4.x from the metadata.json and from the Travis CI testmatrix in January. Our
```
juniorsysadmin: testmatrix

juniorsysadmin:
I have only ever seen one Pull Request actually submitted against a `puppet3` branch.

_posts/2018-12-27-dropping-puppet4.md
```diff
@@ -0,0 +1,22 @@
+---
+layout: post
+title: Dropping Puppet 4 support
+date: 2018-12-27
+github_username: bastelfreak
+---
+
+Puppet 4 will be end of life by the December 31 2018. There isn't an official
```
rjw1: "by December 31st 2018". 
Although I would be tempted to just use the ISO8601 date of 2018-12-31.
"by 2018-12-31"

_posts/2018-12-27-dropping-puppet4.md
```diff
@@ -0,0 +1,22 @@
...
+support matrix for the FOSS components. For Puppet enterprise this can be found
+[here](https://puppet.com/misc/puppet-enterprise-lifecycle). PE 2016.4 is the
+oldest currently support version. It currently ships
```
rjw1: s/support/supported/

oranenj:
I think that spending actual effort on dropping Puppet 4 support is counterproductive.

Puppet is a systems administration tool, and honestly when it comes to infrastructure, things move slowly. I still have to maintain puppet 3 infrastructure because there just isn't sufficient incentive to upgrade, and the effort is significant. Granted, upgrading from 4 to 5 or 6 is much easier, but so is maintaining backwards compatibility.

IMO this EOL date  should just mean that Puppet 4 compatibility is optional going forward. There's no downside to lazily dropping support from modules as features or changes requiring newer versions are introduced. Bump the version and stop testing once such a change occurs, but before that I don't see what it gains us to remove support explicitly.

bastelfreak:
Maybe I can rephrase the initial post a bit. I am fine with dropping support in a lazy way. We could also call it 'Puppet 4 support is optional, it will probably work but we don't test against it'. 

In my opinion, from a users perspective Vox Pupuli delivers FOSS products to users. Users expect that the products work and are secure. I think it it our responsibility to not encourage them to run outdated?EoL software. That not only applies for software we configure, but also for the software that uses our products, like the actual puppet agent.

Except for the email noise from all the PRs, removing it from the testmatrix + updating metadata.json is an easy move, we already have scripts for this by hand and we could do it during one of the upcoming modulesync runs. I agree that we don't need a dedicated patchday for it. Dropping Puppet 4 from the metadata.json in an individual module is easy, removing it from a travis config for an individual module is quite some pain.

ekohl:
I can accept a lazy approach with a target of end of Q1, but we should ask ourselves if we consider this a major version bump. I think it is, but it does have implications for reverse dependencies. We will need to check our other modules if they accept all the latest releases. I have used my own tool (https://github.com/ekohl/puppet_metadata_checker) for this, but it's lacking a README.

Let's do this right and try to improve the ecosystem along the way. I'd propose we make it easy to find cases where any Voxpupuli module doesn't allow the latest released modules and modules that have local changes.

For this we need:
* A way to get a list of all git URLs ([modulesync](https://github.com/voxpupuli/modulesync_config/blob/master/managed_modules.yml)?)
* A way to get a git clone of all these modules (modulesync?)
* A way to check metadata dependencies (https://github.com/ekohl/puppet_metadata_checker)
* A way to check if modules contain unreleased changes

For the last item I don't know what the best approach is. One definition is to use `git log` since the last tag (minus the bump to -RC0) but another is if there's any end user relevant change, which would mean generating a changelog to see if there's any changes. The former is quick, the latter slower and harder to script. That's why I'd propose to start with the former and see how far we get.

jhoblitt:
Under a [current, (current - 1)] support policy, which has been my general assumption for voxpupuli modules, end users have had the time period between the release of puppet 5 and puppet 6 as an effective depreciation notice on support for puppet 4.  I also don't think anyone has suggested removing versions from the forge which support puppet 4.  I'd much rather that PR submitters and maintainers not have to worry about 3 versions and have CI run just a little bit faster.

jhoblitt:
I agree that a major version bump for all modules is reasonable when dropping a major puppet release but I'm not sure how practical that is...

dhollinger:
@jhoblitt `[current, (current - 1)]` is fairly acceptable in my opinion for two main reasons:

1. We are a small community (relatively speaking) and most larger communities (I.E. CentOS) only support `[current, (current - (1|2)]` and even then they generally only support the **latest** patch release. Even vendors that I've worked with have started to adopt the "We can't help you until you patch these bugs".
2. Puppet 4 EoL has been out there since last January (if a user/org is using a tool, but not aware of the lifecycle, that's on them. Not on the vendor or community).

bastelfreak:
@ekohl for some of your items we can already use https://github.com/voxpupuli/modulesync_config/blob/master/bin/count-unreleased-commits and also https://github.com/voxpupuli/modulesync_config/blob/master/bin/get_all_the_diffs. There is also some work going on on tracking reverse dependencies. CC: @binford2k 

A bit related: It might be a good idea to update our website with a dedicated page for puppet (and ruby/facter?) versions we support.

@jhoblitt as a general policy `[current, (current - 1)]` sounds good to me. We are currently in the process of getting more travis slots and also optimizing the pipelines, to speed up CI. Also dropping puppet 4 would speed this up.

dhollinger:
@oranenj I'm fine with lazy removal of Puppet 4 from modules (no need to forcefully do it), but disagree on maintaining testing/support/patches for Puppet 4 (or even Puppet 3). 

I still stand by the idea that by end of Q1 2019, Puppet 4 be removed from the metadata.json and we stop testing against Puppet 4. We have resource issues on travis as it is and are too small to continue supporting Puppet 4 across our modules. We don't need to add code that breaks Puppet 4, I am absolutely against that.  I DO think by end of Q1 2019, if a user/company/org wants to continue using the module on an EoL version of Puppet, VP members should not be obligated (I use this term very loosely) to support the code on that version (outside of helping with a user's upgrade process).

suckatrash:
> @jhoblitt `[current, (current - 1)]` is fairly acceptable in my opinion for two main reasons:
> 
> That said, I don't think we should write Puppet 4 compatibility out via code, but drop it from `metadata.json` by end of Q1 2019. I also think we should drop patch support for Puppet 4-specific patches starting 2019/1/1

As someone who supports PE I think this approach makes sense and is reasonable.  I don't see any reason to rip out Puppet 4 functionality.  If someone needs that functionality they can always use an older module version with the understanding that it's EOL.  Saying that Puppet 4 is no longer developed / tested on and removing from metadata should be sufficient.

When Support is contacted by PE customers for issues with EOL versions, we provide a best-effort attempt at resolving the issue, but our main focus is often understanding what the blockers are to getting the customer upgraded and back onto a supported path (and making sure they have a plan to do so).

Dan33l:
If we set a hard limit like 2019-01-01 or 2019-03-31, we are under pressure. What is happening if the process is not finished on 2019/03/31 ? IMO we just need to post that decommissioning process of Puppet 4 is beginning.

After this, we need to know the path we want/can use to make this task.

I also disagree to create a branch `puppet4`. Tag is the already existing tool.

LongLiveCHIEF:
Not commenting on the PR per se, but an observation as someone who has spent time with an organization with dozens of puppet developers.

Puppet has been moving pretty quickly the last 2 or 3 years, but the education and docs has not kept pace. It's difficult for developers to keep up, when essential tools (testing puppet for example) have very limited docs and education.

We just need to keep that in mind as we mull decisions like this. In many other languages and frameworks, this is not a problem and therefore the community doesn't have to worry about things like technical debt in modular OSS.

_posts/2018-12-27-dropping-puppet4.md
```diff
@@ -0,0 +1,22 @@
+---
+layout: post
+title: Dropping Puppet 4 support
+date: 2018-12-27
+github_username: bastelfreak
+---
+
+Puppet 4 will be end of life by the December 31 2018. There isn't an official
+support matrix for the FOSS components. For Puppet enterprise this can be found
```
ahpook: sorry, the page used to have a statement about this but i think it got lost in a rewrite. I've added this block now, which hopefully clarifies the situation:

> Puppet Inc's open-source projects which are upstream of Puppet Enterprise (such as Puppet and Facter) track the maintenance and EOL dates for the PE release streams which depend on them. For example, the PE2018.1 LTS uses Puppet Agent 5.5, so the open source Puppet Agent 5.5 will receive bug fix and security updates until May 2020. For a full list of upstream components in PE releases, see the "Component Versions in Recent PE Releases" page on the docs site.

ragnarkon:
Also agree with lazily dropping support. I don't think setting a strict timeline ("it will be done by end of Q1", for example) is strictly necessary. To me it seems easiest to remove Puppet 4 from the metadata.json during the next major (or even minor) release. However, some organizations will appreciate having a timeline they can follow.

Regardless of the exact timeline/method, it doesn't make any sense to put effort put maintaining Puppet 4 support. Something that has been a *huge* struggle for me in my organization is getting everyone to keep up with latest changes. The staging module, for example, has been deprecated for over a year now and I'm still trying to get people to move over to the archive module.

It is already hard enough to get people to keep their modules somewhat current, and I certainly don't want the vendor or the community giving folks more excuses to ignore currency on their modules. Having support for Puppet 4 when Puppet itself doesn't support Puppet 4 is really just allowing complacency to spread within my organization (and I suspect others as well).  As @dhollinger said earlier, we really should be moving folks towards agile practices, not giving them excuses to avoid them.

_posts/2018-12-27-dropping-puppet4.md
bastelfreak: hey, I'm not quite sure what you mean. Is there a typo I cannot see?

_posts/2018-12-27-dropping-puppet4.md
bastelfreak: fixed, thanks!

_posts/2018-12-27-dropping-puppet4.md
bastelfreak: thanks, fixed it

bastelfreak:
Hey everybody. Thank you so much for the detailed responses.

@ekohl I never had the intention to do >100 releases just to release this particular change. Rather I won't to follow the usual "we release when we've time or a contributor requests it" schedule. I added this phrase:

> We won't do releases just for this change, but rather follow our normal release schedule.

####################

I updated the blog post. Does this know match the expectations from the majority of the people?

bastelfreak:
ah good catch :D I will update it.

_posts/2019-01-03-dropping-puppet4.md
```diff
@@ -0,0 +1,23 @@
+---
+layout: post
+title: Dropping Puppet 4 support
+date: 2019-01-03
+github_username: bastelfreak
+twitter_username: BastelsBlog
+---
+
+Puppet 4 is end of life since 2018-12-31. There isn't an official
```
dhollinger: `Puppet 4 reached End of Life on 2018-12-31.` Sounds a bit better in English.

jhoblitt: How about

> Me remembers it like it was yesterday. In December 31 of 2018, when Puppet 4 walked the plank.

or

> Well shave me belly with a rusty razor! Puppet 4 is end of life since 2018-12-31

?

_posts/2019-01-03-dropping-puppet4.md
```diff
@@ -0,0 +1,23 @@
+---
+layout: post
+title: Dropping Puppet 4 support
+date: 2019-01-03
+github_username: bastelfreak
+twitter_username: BastelsBlog
+---
+
+Puppet 4 reached End of Life since 2018-12-31. There isn't an official support
```
ekohl: Not a native speaker, but I think it's "reached ... on" or "is ... since". This combination looks odd to me.

bastelfreak: ugh, I made many typos here. Fixed it.

dhollinger:
LGTM 👍

PR #169, opened 2019-01-14
added hiera v5 plugins

maxadamo:
I have added a a 3rd column in the Hiera table only 
The first column shows the plugin for hiera v3, the second column shows the plugin link for hiera v5, the 3rd column shows the description.


bastelfreak:
HI @maxadamo, thanks for the PR! At the moment we require commits that are signed with gpg in this repo. Are you able to set this up?

maxadamo:
Ok. If you go to the commits tab it shows a second commit signed with GPG.

bastelfreak:
thanks! Could you squash your commits together so both are signed?

PR #189, opened 2019-06-10
update docs for data-in-modules

bastelfreak:
CC: @ekohl


traylenator:
Makes sense, it's much easier to read the manifest with defaults in `init.pp`

PR #192, opened 2019-07-11
Add more advice on PR merging

alexjfisher:


docs/index.md
```diff
@@ -179,6 +179,8 @@ There are a few things that can be checked if you review a pull request against
 
 * You can merge your own PR if it was approved by a collaborator with merge permissions and travis is green. Don't merge if either one of those conditions are not true
   * Modulesync PRs are an exception (a PR based on changes that the msync tool did, NOT PRs on [modulesync_config](https://github.com/voxpupuli/modulesync_config#modulesync-configs)). We agreed some time ago that it's ok to merge your own modulesync PR if travis is green, without separate approval. This is okay because changes to [modulesync_config](https://github.com/voxpupuli/modulesync_config#modulesync-configs) were reviewed and tested
+  * If your PR is non-trivial or perhaps has only been approved by a work colleague etc, please consider allowing reasonable extra time for other 3rd parties to leave their reviews before merging.
```
binford2k: How about something like this? Trying to evoke the idea of reviews by people who might see the issue from different angles rather than just numeric count.

> If the PR is non-trivial or hasn't been reviewed by many different kinds of people, then please consider allowing reasonable extra time for other interested or affected parties to leave their reviews before merging.

ghoneycutt: @binford2k that's a great idea but puts a lot of pressure on the person sending the PR. They have to know who people are and what their viewpoints might be, which is not realistic.

Issue #195, opened 2019-08-03
license link is 404


ghoneycutt:
https://voxpupuli.org/elections/

at the very bottom right there is a link for `Dual license (CC BY-SA 4.0 + Apache 2.0)` which points to `https://voxpupuli.org/elections/license/` which is a 404.

bastelfreak:
This is fixed now
![2019-08-04-111720_739x58_scrot](https://user-images.githubusercontent.com/341866/62421869-7038a700-b6a9-11e9-97f6-d616b959fc05.png)

PR #196, opened 2019-08-03
(GH-195) Fix license link 404

ghoneycutt:
Without this patch, the link is relative and works on the home page and
404's everywhere else.


ghoneycutt:
Fixes #195

bastelfreak:
Thanks for the fix @ghoneycutt !

PR #204, opened 2019-08-09
The start of a set of reference architectures

binford2k:
Do not merge this as is. It's not ready to go live on the site.
You'll want to use the `Checkout via command line` instructions on the
PR to check this out locally so you can tweak the graph and see it live.

Check out the PR, run the site locally with `jekyll serve`, tweak and
improve as needed.

### Note that this is not intended to be an actual suggested architecture!

It's a reasonable example, for sure, but the purpose of this PR is
simply to create the framework and demonstrate an example so that
you can more easily put together the reference yourself.

See https://mermaidjs.github.io/#/flowchart for syntax documentation.
Also, if anyone knows of a better text based chart DSL, then please suggest.

Screen Shot 2019-08-09 at 2 17 33 PM

### DIFF
--- a/_docs/master_agent.md
+++ b/_docs/master_agent.md
@@ -1,0 +1,109 @@
+---
+layout: architecture
+title: Master of Masters
+date: 2019-07-09
+version: v0.0.1
+summary: A complete master/agent architecture with multiple compile masters and load balancing for redundancy.
+---
+
+<div class="mermaid">
+  graph LR;
+  git(Git Repository)
+  Foreman(The Foreman)
+  Webhook(Puppet Webhook Server)
+  AllCM((All Compile Masters))
+
+  PuppetDB
+
+  PuppetServerMoM{Master of Masters}
+
+  subgraph CM[Compile Masters]
+      Compile1[/Compile Master\]
+      Compile2[/Compile Master\]
+      Compile3[/Compile Master\]
+  end
+
+  LoadBalancer[Load Balancer]
+
+  Agent1(Agent 1)
+  Agent2(Agent 2)
+  Agents(Agents ...)
+  Agent_n(Agent n)
+
+  click Foreman "https://www.theforeman.org" "Foreman is a complete lifecycle management tool for physical and virtual servers."
+  click Webhook "https://github.com/voxpupuli/puppet_webhook" "A webhook service that can trigger code deploys from source code repository updates."
+
+  git --webhook--> Webhook
+  Webhook --r10k code deploy--> PuppetServerMoM
+  Webhook -.r10k code deploy.-> AllCM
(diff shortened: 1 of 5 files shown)


ekohl:
That looks nice!

ekohl:
Do you intend to add some guides on deployment to this? Where do you want to take this?

ghoneycutt:
This might benefit from having an attended audience section and where this might be used.

binford2k:
My goal is pretty simple. I just want to encourage a little more consistency in the OSS Puppet world. That said, I don't think that Puppet, Inc should have too much influence of the actual reference. I'm happy to facilitate in every way, but I think the actual recommendations should come from the community. That's why the giant disclaimer in the PR.

A secondary goal is to encourage another way to get involved with Vox Pupuli. I'd love to get community members contributing to a shared knowledge repository and Vox is the de facto group for community experts.

I'll start noodling on an audience & usage section and I think that some deployment guides would be amazing too.

_docs/master_agent.md
```diff
@@ -0,0 +1,120 @@
...
+
+[Foreman](https://www.theforeman.org) is a complete lifecycle management tool
+for physical and virtual servers. It will provide you with a graphical
+classifier, a Hiera data source, and report monitoring. It also includes the
+power to easily automate repetitive tasks, quickly deploy applications, and
+proactively manage servers, on-premise or in the cloud.
+
+
+### Puppet Webhook
+
```
Dan33l: In this section we probably should point to puppet module [puppetwebhook](https://github.com/voxpupuli/puppet-puppetwebhook/.)

_docs/monolithic.md
```diff
@@ -0,0 +1,97 @@
...
+
+[Foreman](https://www.theforeman.org) is a complete lifecycle management tool
+for physical and virtual servers. It will provide you with a graphical
+classifier, a Hiera data source, and report monitoring. It also includes the
+power to easily automate repetitive tasks, quickly deploy applications, and
+proactively manage servers, on-premise or in the cloud.
+
+
+### Puppet Webhook
+
```
Dan33l: In this section we probably should point to puppet module [puppetwebhook](https://github.com/voxpupuli/puppet-puppetwebhook/.)

rnelson0:
Any chance this can get committed? I don't think it needs to be perfect to start out with, but it would be good to have it live and see it in action before making further modifications.

bastelfreak:
I think we should finally merge it and update it later on

dhoppe:
@binford2k I think we should not use our own module to manage the deployment of the Puppet server. As far as I know, this module is not longer under maintenance and people should use the Foreman module.

PR #202, opened 2019-08-09
This breaks the monolithic docs page up

binford2k:
It creates a section of smaller pages, with a nice list on the index
page. It also adds a simple breadcrumb to make navigating back out a bit
simpler. It doesn't change actual content.

Screen Shot 2019-08-08 at 5 52 26 PM

PR #210, opened 2019-10-05
Add missing links

baurmatt:
Seems like those links gone missing in #202.


bastelfreak:
Hi @baurmatt, thanks for providing this fix. This repository requires commits signed with gpg. Can you please do that for your change?

baurmatt:
Hey @bastelfreak, sorry missed that. Just signed and force pushed it again.

PR #212, opened 2019-11-28
update release docs for github token

bastelfreak:


_docs/releasing_version.md
```diff
@@ -32,6 +32,9 @@ bundle install --path .vendor/ --without system_tests development; bundle update
 ```
 
 We can generate the changelog after updating the metadata.json with a rake task:
```
ekohl: The semi colon at the end feels a bit weird. You can move the docs about the token till after the exact command and provide an example with `CHANGELOG_GITHUB_TOKEN="mytoken" bundle exec rake changelog`

PR #213, opened 2019-11-29
Add page on archiving modules

alexjfisher:


_docs/archived_modules.md
```diff
@@ -0,0 +1,15 @@
...
+---
+
+It is Vox Pupuli's mission to provide a home for modules so that development can continue after the original authors/maintainers have moved on.
+In rare circumstances a module may be archived and placed into a read-only state.  This typically will only occur if the module/project no longer serves a valid
+purpose.  For example, a module that interacted with a discontinued 3rd party service.
+
+If you come across an archived Vox Pupuli project that you are still using or think should not have been archived, [let us know][email].
+The decision to archive a module isn't necessarily final and can be undone.
+
+[email]: https://help.github.com/en/articles/generating-a-new-gpg-key
```
ekohl: Looks like an incorrect URL for this

alexjfisher: Oops

alexjfisher: Fixed (and renamed to accommodate policy on deprecations etc)

_docs/deprecated_and_archived_modules.md
```diff
@@ -0,0 +1,17 @@
...
+layout: post
+title: Deprecated and Archived Modules
+date: 2019-11-29
+summary: Vox Pupuli policy on deprecating and archiving modules
+---
+
+## Archived Modules
+
+It is Vox Pupuli's mission to provide a home for modules so that development can continue after the original authors/maintainers have moved on.
+In rare circumstances a module may be archived and placed into a read-only state.  This typically will only occur if the module/project no longer serves a valid
```
Dan33l: Rare or not it not the focus. We can archive a module as often as required. A simple `But` LGTM.

alexjfisher: I strongly believe it should be a last resort (and thus rare).  It's not fair on those who have migrated modules to us to archive them without *very* good reason.

bastelfreak:
@alexjfisher can you please rebase?

bastelfreak:
@alexjfisher ping :)

bastelfreak:
mhm I cannot push to your branch :sadface:

bastelfreak:
should we update this a bit? If we archive a module I expect that we check if another successor is available. If so, this needs to be mentioned on forge.puppet.com. In all cases the module should be deprecated on the forge and it would be nice if the README.md also contains a notice with deprecation/successor notice.

_docs/deprecated_and_archived_modules.md
```diff
@@ -0,0 +1,17 @@
+---
+layout: post
+title: Deprecated and Archived Modules
+date: 2019-11-29
```
ekohl, line 4: Will people see this post if it's so far in the past?

bastelfreak: it's not in the blog section: https://voxpupuli.org/docs/

PR #215, opened 2020-02-05
Don't require breaking releases for legacy OS

bastelfreak:


_docs/reviewing_pr.md
```diff
@@ -16,7 +16,10 @@ one of our modules:
 * Were changes to master merged that are required in this PR (for example an updated Travis-CI configuration)? Add the `needs-rebase` label
 * Does it need additional tests? Add the `needs-tests` label
 * Does it have failing tests? Add the `tests-fail` label
-* Does it drop support for a specific Operating system or a major Puppet version? Add the `backwards-incompatible` label
+* Does it drop support for a specific Operating system?
```
ekohl:
```suggestion
* Does it drop support for a specific Operating system version?
```

_docs/reviewing_pr.md
```diff
@@ -16,7 +16,10 @@ one of our modules:
 * Were changes to master merged that are required in this PR (for example an updated Travis-CI configuration)? Add the `needs-rebase` label
 * Does it need additional tests? Add the `needs-tests` label
 * Does it have failing tests? Add the `tests-fail` label
-* Does it drop support for a specific Operating system or a major Puppet version? Add the `backwards-incompatible` label
+* Does it drop support for a specific Operating system version?
+  * Does it also drop code for that OS? Add the `backwards-incompatible` label
```
ghoneycutt: If the platform is end of life (EOL) and no longer supported, then this is fine. There should be a tagged released before doing this and the only change should be the removal of this platform.

If the platform is not EOL, this should be tagged `backwards-incompatible`.

bastelfreak:
@ekohl can you approve please?

_docs/reviewing_pr.md
alexjfisher: That sounds reasonable to me.  Perhaps this belongs in a separate section of the docs though.  Seems quite hard to capture this in bullet point format under general PR review guidelines.

ghoneycutt: Here's what we do with the Sensu module

https://github.com/sensu/sensu-puppet#notes-regarding-support

llowder: I have mixed feelings on this.  As a module maintainer, it makes sense, but as a module consumer who has to support older OSes, even sometimes after they are officially EOL, having that major version bump is simpler for knowing which versions I can use (though in some cases, the removal is in name only and doesn't actually mean it won't work on those OSes). And while yes, each update should be reviewed, minor / patch level changes don't normally require the same caution as majors, so this is something that could be missed.

Rather than an all or nothing type view, my thoughts are that if the dropping of support is just a change to metadata, then maybe a major isn't needed. But if the dropping of support requires code changes at all (for example, removing defaults for the old OS from params.pp or such), then it should be major.

ghoneycutt: @llowder As someone who supports ancient versions of Suse and Solaris, I totally get this sentiment. I don't think it is right for VP though. Removing old, EOL platforms quickly makes it easier to maintain the modules. That balanced with a final release for a platform gives consumers the ability to pin at that version.

The only thing changing would be metadata and code for an EOL platform and making that a major version bump would ask, what other things should be changed while we are doing the major version bump and what other backward incompatible changes are in the queue? This in practice leads to PR's with the backward incompatible tag just sitting around forever and people hesitant to do a major version release. I believe it slows down our efforts as a whole.

If we need to do a major version bump to software to remove EOL platforms, it will go slowly if at all and cause more harm than good.

ekohl: I'll echo what @llowder said. What if someone drops the version from metadata.json because it's EOL, but not the code. Later someone comes along and and thinks it can be cleaned up because it's not mentioned in metadata.json and thus unused. That means twice you do a minor or even patch version for a breaking change. Keeping the rule "code remove -> API change -> major version" is safe and doesn't require too much thinking. Overall we haven't been shy to do major versions so I don't see that as a huge concern.

ekohl:
This now needs a rebase since it's out of data. Other than that I'd like to move forward with this. I ran into it with https://github.com/voxpupuli/puppet-trusted_ca/pull/25 where there are pretty much no facts used and it was just an adminstrative thing (and wasting CI cycles).

_docs/reviewing_pr.md
dhollinger: I'm with @llowder and @ekohl, though a part of me would rather just keep removal of EOL OS support as a major change. The major change spec, from how I read it, is for any change (functional or non-functional) that could be perceived as "breaking" to a user and a metadatqa change could be exactly that.

dhoppe:
I think we should do a major release and then modify the ```metadata.json``` and the relevant code. 

We should also add a section to ```README.md``` that describes which version of the module should be used if someone needs support for an obsolete release. I have seen something like that for other modules.

bastelfreak:
this is now obsolete with https://github.com/voxpupuli/voxpupuli.github.io/pull/232

PR #216, opened 2020-03-02
Run Puppet Strings before releasing a new version

baurmatt:
@bastelfreak suggested on my last release PR for puppet-prometheus to regenerate the Puppet Strings documentation because it was outdated. I think this is a good thing to do on all releases and should therefore be mentioned in our documentation.

### DIFF
--- a/_docs/releasing_version.md
+++ b/_docs/releasing_version.md
@@ -31,6 +31,15 @@ And in case you installed the gems before:
 bundle install --path .vendor/ --without system_tests development; bundle update; bundle clean
 ```
 
+If the module contains a Puppet Strings generated documentation, please
+ensure the file is update to date. A good indicator for a Puppet Strings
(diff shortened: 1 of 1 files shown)


_docs/releasing_version.md
```diff
@@ -31,6 +31,15 @@ And in case you installed the gems before:
 bundle install --path .vendor/ --without system_tests development; bundle update; bundle clean
 ```
 
+If the module contains a Puppet Strings generated documentation, please
+ensure the file is update to date. A good indicator for a Puppet Strings
+documentation is the existence of a REFERENCE.md file. You can automatically
+generate the documentation by running the following rake task:
+
+```bash
+bundle exec rake strings:generate:reference
```
ekohl: We also have a task `reference` which at least in puppet-extlib is used to set the patterns, though it's not needed right now. Perhaps we should just remove that and enhance the task if needed.

PR #222, opened 2020-08-09
clarify data-in-modules usage

bastelfreak:


_docs/reviewing_pr.md
```diff
@@ -21,7 +21,7 @@ one of our modules:
...
 * Are datatypes from stdlib used? Ensure that lowest supported stdlib version is 4.18.0 (This is the first version that supports Puppet 5). Check if a newer version introduced the used datatype
-* Are hiera yaml files added for data-in-modules? Ensure that the data is compatible with [hiera 5](https://puppet.com/docs/puppet/5.3/hiera_migrate.html#use-cases-for-upgrading-to-hiera-5). Static data that is equal across every supported operating system can stay in the init.pp, it doesn't need to be moved to a common.yaml
+* Are hiera yaml files added for data-in-modules? Ensure that the data is compatible with [hiera 5](https://puppet.com/docs/puppet/5.3/hiera_migrate.html#use-cases-for-upgrading-to-hiera-5). Static data that is equal across every supported operating system must stay in the init.pp, it doesn't need to be moved to a common.yaml (until [puppet-strings issue #250](https://github.com/puppetlabs/puppet-strings/issues/250) is solved). Data that differs between the hierachies must have no default value in a commons.yaml/default.yaml. Otherwise people can assume that it doesn't differ.
```
alexjfisher, line 24: >  Data that differs between the hierachies must have no default value in a commons.yaml/default.yaml.

So what are you saying should go in `default.yaml`?

bastelfreak: basically nothing at the moment. I tried to write down what was the result of a few IRC/slack discussions. Putting static data into a default.yaml is bad because puppet-strings cannot work with it and assumes the parameter has no default value. If we add data to it that differs based on the operating system it implies that we also add the same key to something like os/RedHat-8.yaml. But people that see the data in default.yaml assume that this is the *default* for everything.

In my opinion default.yaml should only be used for static data if https://github.com/puppetlabs/puppet-strings/issues/250 is solved. And it should not be used for anything else.

ekohl: I'm not entirely sure what the benefit of a `default.yaml` is over putting it in `init.pp` (or any other file).

A big downside is that you now have to look in two places for something. If you remove a parameter, there's a greater chance you forget to remove parts of it. You may argue that you need to check data files anyway, but minimizing sprawl over multiple files is a good goal to keep.

The only thing I can see that would benefit from moving is a large default, like a list of ciphers. Those can break up readability.

I also don't see https://github.com/puppetlabs/puppet-strings/issues/250 being resolved any time soon. Let's not speculate about the future and keep instructions short & simple.

```suggestion
* Are hiera yaml files added for data-in-modules? Ensure that the data is compatible with [hiera 5](https://puppet.com/docs/puppet/5.3/hiera_migrate.html#use-cases-for-upgrading-to-hiera-5). Static data that is equal across every supported operating system must stay in the init.pp, it shouldn't be moved to a common.yaml due to [puppet-strings issue #250](https://github.com/puppetlabs/puppet-strings/issues/250).
```

I removed the part about not being in `common.yaml` because I think you can put something in `common.yaml` that you then only override for EL8. I see it as:

* All the same -> init.pp
* All the same except some -> common.yaml + overrides
* No common value -> overrides

bastelfreak:
I merged the suggestion from @ekohl . Please have a look again. If there isn't a majority against this in the next days, I'm going to merge this PR.

PR #227, opened 2020-08-27
Update architectures to use more inclusive terminology

binford2k:


_docs/arch_load_balanced.md
```diff
@@ -1,32 +1,32 @@
...
   git(Git Repository)
   Foreman(The Foreman)
   Webhook(Puppet Webhook Server)
-  AllCM((All Compile Masters))
+  AllCM((All Compilers))
 
   PuppetDB
 
-  PuppetServerMoM{Master of Masters}
+  PuppetServerMoM{Main Puppet Server}
```
ghoneycutt, line 24: Awesome work @binford2k 

Noticed that PuppetServerMoM is still there, perhaps `s/MoM/MPS` ? Anyhow, it may leave people scratching their head what a Puppet MoM is. Simlar thing with CM below.

PR #228, opened 2020-11-17
Mention Puppet5/RHEL6 deprecation

bastelfreak:


wyardley (approved):

anarcat:
wow, really? we're going to just totally remove puppet 5 support everywhere already? some folks have barely managed to upgrade now... and Debian still doesn't have Puppet 6 packaged...

this feels a little rushed. i understand if we'd deprecate it, but actively removing it seems rather harsh...

bastelfreak:
The first puppet 6 release was more than two years ago. How much time should Puppet Inc. provide to people to upgrade? I don't think it is a good idea to keep supporting a puppet version in our modules that is dead upstream and won't receive any updates. We should not encourage people to use outdated software.
Re: Debian: To be fair, basically every package in Debian 10 is old. And if we keep supporting it we just encourage people to run this old stuff even longer? As a workaround, Puppet provides their own repos for Debian.

Most of our modules don't have Puppet 5 specific code, also nobody is forced to upgrade (and why would you want to run the latest module version but a 1,5 years old Puppet release).

logicminds:
Worst case scenario here is the user pins to specific versions of the module or gems.  Removing support won't affect them.

anarcat:
> How much time should Puppet Inc. provide to people to upgrade?

5 to 12 years, depending on how committed you are to your users. :)

But hey, it's your call, you just asked for a review from everyone, and I'm telling you: I am far from upgraded to 6. Snipes about how "old" Debian is and how "we need to avoid outdated software" are not useful to me... 

I know that Puppetlabs provides Debian packages themselves, but those have strings attached to them (e.g. they do not follow Debian policy). There are reasons why people struggle to maintain that code inside Debian (e.g. trust path).

Anyways, all I was proposing was to provide some deprecation notice instead of just actively removing support. If that's not current policy or too much work, I understand. ;)

roidelapluie (approved):
:heavy_check_mark:

igalic:
> The first puppet 6 release was more than two years ago. How much time should Puppet Inc. provide to people to upgrade? I don't think it is a good idea to keep supporting a puppet version in our modules that is dead upstream and won't receive any updates. We should not encourage people to use outdated software.
> Re: Debian: To be fair, basically every package in Debian 10 is old. And if we keep supporting it we just encourage people to run this old stuff even longer? As a workaround, Puppet provides their own repos for Debian.
> 
> Most of our modules don't have Puppet 5 specific code, also nobody is forced to upgrade (and why would you want to run the latest module version but a 1,5 years old Puppet release).

@bastelfreak this ⬆️ feels like it belongs into the blog post itself

alexjfisher:
I think this is unnecessarily aggressive.  Stopping testing against puppet 5 is probably fine, but I'm not ok with actively removing EL6 code (or puppet 5 code) until such time that supporting it in a given module is actually painful.  Doing so, alienates a large number of users who might otherwise submit and review PRs.

I also don't want to preach to users, (telling them we're doing them a favour by telling them how old their systems are and that us dropping stuff is for their own good).

_posts/2020-11-17-dropping-puppet-5-and-RHEL-6-support.md
```diff
@@ -0,0 +1,13 @@
+---
+layout: post
+title: Dropping Puppet 5 and RHEL 6 support
+date: 2020-03-29
```
alexjfisher, line 4: Incorrect date.

wyardley:
> I also don't want to preach to users, (telling them we're doing them a favour by telling them how old their systems are and that us dropping stuff is for their own good).

Agree with all of this. As a volunteer effort, it's reasonable that Voxpupuli has to be strategic about what to support and how long to support things for, however, also agree that officially supporting / testing versions is one thing, but going out of our way to make it harder for people to use things that probably would work is less useful.

> > (and why would you want to run the latest module version but a 1,5 years old Puppet release).

> because you need a bug fix / feature that the new module has

had this exact situation a lot when I was working with Puppet, so I agree that this is pretty common.

ekohl:
> wow, really? we're going to just totally remove puppet 5 support everywhere already? some folks have barely managed to upgrade now... and Debian still doesn't have Puppet 6 packaged...

That does remind me: my personal infra is also using Puppet 5 because I prefer not to use the AIO packages. It looks like Debian Testing also only has Puppet 5. That should be a priority as well.

jhoblitt:
In case folks aren't aware, RHEL6/centos 6 is EOL in ~13 days.

ghoneycutt:
@alexjfisher Check out https://github.com/ghoneycutt/puppet-module-pam/pull/232/commits/3c29af84615545ab51befe8ac340903754ff48b7 where Debian-8 (EOL) is being gently decommissioned.

In my pam module when something is EOL it no longer has acceptance tests run for it and the README describes it as it probably works but we don't care too much anymore :) Since a lot of work goes into supporting any platform, we keep the spec/fixtures and unit tests around. 

Well actually the readme has the following with a list platforms From https://github.com/ghoneycutt/puppet-module-pam#may-work

```
### May work
These platforms have spec tests and have been verified in the past,
though are not functionally tested and formally supported.
```

This will make sense on a module by module basis, so you might not rip out EL6 bits but we definitely don't want to waste test cycles and trying to debug EOL containers/VM's.

If there are a lot of conditionals around EL6 in a module, which adds complexity, well, someone may rip it out and that should be OK. Removing it just for sake of doing it when not part of another refactor or the like should be discouraged.

_posts/2020-11-17-dropping-puppet-5-and-RHEL-6-support.md
```diff
@@ -0,0 +1,13 @@
...
+github_username: bastelfreak
+twitter_username: BastelsBlog
+---
+
+Hi Vox Pupuli followers!
+Puppet 5 will be end of life in the next days, Puppet 7 is around the corner.
+Also RHEL/CentOS will be end of life in the next few days. In the next few days
```
ghoneycutt, line 11: instead of repeating in the next few days, suggest listing the dates that each is EOL and stating that after that date, support will drop as well as functional testing.

Fabian1976, line 11: All RHEL/CentOS will be end of life?

But agree on the support dropping. If people still haven't upgraded form RHEL6 or Puppet5, they won't update their modules either, so they won't be affected by this.

alexjfisher: > If people still haven't upgraded form RHEL6 or Puppet5, they won't update their modules either

From experience in multiple companies over the last 5 years, I can guarantee that's exactly what people do.  As @igalic said, "upgrading a module is infinitely easier than upgrading a configuration management system"

ekohl:
> In case folks aren't aware, RHEL6/centos 6 is EOL in ~13 days.

This isn't entirely true. CentOS 6 is EOL on November 30th and so is RHEL standard support. Customers can buy Extended Life Cycle Support (ELS) for RHEL 6 which is supported till June 30, 2024. See https://access.redhat.com/articles/4665701.

I'm not saying we should stick to that, but we should at least keep the facts straight :)

roidelapluie:
Yeah, technically CentOS is not supported at all.

oranenj:
I also have to note that I have a couple systems where it's not *possible* to upgrade from Puppet 5; they're Debian hosts running on ARM, and Puppet doesn't provide upstream packages for those.

A soft deprecation is fine, but there are still valid reasons to continue supporting Puppet 5

jhoblitt:
@ekohl You can also buy extended support for rhel4... that doesn't necessarily mean that the entire puppet community should have to support it.  In fact, I haven't seen a module that actually tests with RHEL... everyone I've seen is using centos or very rarely SL.

ekohl:
@jhoblitt note my last line. I just wanted to make it clear that it's still supported in some cases. That means someone might care about support in the modules. If it's not a large support burden (=a lot of additional logic) then IMHO it's fair for a module to choose to support it.

I generally don't like the soft support. The "it may work, but we don't test it" feels like a weak situation. Especially on non-trivial modules it's easy to break something.

b4ldr:
Like others i understand the practicality of trying to keep pase with puppet support however i feel that while major distributions like debian are still using puppet 5, puppet 5 is still worth supporting.  Otherwise we risk someone forking (and all the issues that brings) just to maintain puppet 5 compatibility.

jhoblitt:
@ekohl I'm not a fan of soft support either.  I am a fan of explicitly removing support for unsupported platforms as it makes the code base smaller and easier to refactor down the road.  If someone really has a good reason for using rhel3 and puppet 2.7.0 then they hold onto the same versions of puppet modules they were using in 2011 too.

jhoblitt:
Can we split the discussion for rhel6 and puppet5.  I am in favor of dropping rhel6 but I think it is completely reasonable to hold onto puppet5 at least until 7 ships.

baurmatt:
@jhoblitt Puppet 7 is already tagged in the Github repository -> https://github.com/puppetlabs/puppet/releases/tag/7.0.0

I'm also in favor of deprecation but not actively removing support if there isn't a specific reason (failing tests/special logic/...)

ekohl:
:+1: on splitting the discussion. Other than both being EOL, they're quite different.

dobbymoodge:
@ekohl Shouldn't this PR be closed, as it's superceded by #229 and #230?

ekohl:
Yes, let's close it and continue the discussion in the separate issues.

PR #232, opened 2020-12-09
Add documentation on support lifecycle

dhollinger:


_docs/software_support_lifecycle.md
```diff
@@ -0,0 +1,78 @@
...
+summary: Vox Pupuli policy on supporting upstream software
+---
+
+## Software Support Lifecycle
+
+As a part of Vox Pupuli's mission to provide a home for modules that still have a userbase and need development, we need to define a clear support strategy for what operating systems, dependencies, and upstream software we will support and for how long those will be supported.
+
+As we operate entirely on the graces of our community members being willing to voluntarily contribute to our modules, we also need to be respective of their time and resources and not add undue stress in order to support software that upstream creators no longer support.
+
+To account for this, we no longer support any software or operating system after the date that the maintainers have marked said software or operating system as End of Life. However, this does not mean we will immediately remove code that allows for the module to run on End of Life'd software or operating systems, but the community will not longer be expected to provide support to users for that software.
```
gcoxmoz, line 14: typo, I think: 'community will NO longer'

_docs/software_support_lifecycle.md
```diff
@@ -0,0 +1,78 @@
...
+
+There are exceptions and the full policy is detailed in the next section.
+
+### Support Lifecycle
+
+#### Operating Systems
+Operating System support is considered no longer supported once it reaches an End of Life date determined by the upstream maintainer* of the operating system. To ensure our users have enough time to adjust, we will take the following steps to ensure as seamless of a transition as possible:
```
gcoxmoz, line 21: Maybe s/Operating System support is considered/An Operating System is considered/

_docs/software_support_lifecycle.md
```diff
@@ -0,0 +1,78 @@
...
+
+1. A deprecation PR will be created no later that 3 months before an Operating System's End of Life date.
+2. Removal of the End of Life version from the metadata.json and tests to be completed by the End of Life date.
+3. Issues relating to the End of Life operating system will will no longer be addressed after the End of Life date.
```
gcoxmoz, line 25: s/will will/will/

_docs/software_support_lifecycle.md
dhollinger: Oh, goodness I really messed that first sentence up. Nice catch

_docs/software_support_lifecycle.md
```diff
@@ -0,0 +1,78 @@
...
+4. Code allowing a module to be run on the End of Life operating system will be left in the module until one of the following milestones are reached after which time the code will be removed:
+    1. Maintaining the code prevents the module from moving forward.
+    2. 12 Months after the End of Life date.
+5. Pull Requests will continue to be accepted with the following exceptions:
+    1. Accepting such a pull request would be detrimental to the module's long term goals.
+    2. The code has already been removed as per the standards set in section 4.
+
+##### Support Table:
```
gcoxmoz, line 35: Suggestions: YYYY-MM-DD dates, for more uniform readability.  And, link to where you got the date.  Say, hypothetically, in case someone decides to move dates on you.

dhollinger: Yeah, it was a pain to find all those dates. I think I'll follow @binford2k's suggestion and link to https://endoflife.software/operating-systems for most of them.

For some reason, SUSE and OpenSUSE are not on that site though

_docs/software_support_lifecycle.md
```diff
@@ -0,0 +1,80 @@
...
+
+*maintainers include, but are not limited to, the community, company, or organization that maintains ownership over a specific piece of software or project.
+
+##### Support Table:
+
+| OS Name | Support Begins | End of Life | Notification PR | End of Support | Removal of Code |
+| ------- | -------------- | ----------- | --------------- | -------------- | --------------- |
+| CentOS/RHEL 6 | July 10th, 2011 | Nov 30th, 2020 | N/A | Nov 30th, 2020 | Nov 30th, 2021 |
+| CentOS/RHEL 7 | July 7th, 2014 | June 30th, 2024 | Mar 31st, 2024 | June 30th, 2024 | June 30th, 2025 |
+| CentOS/RHEL 8 | Sept 24th, 2019 | Dec 31, 2021 | Sept 30th, 2021 | Dec 31, 2021 | Dec 31, 2022 |
```
alexjfisher, line 41: RHEL 8 of course won't be EOL next year.

_docs/software_support_lifecycle.md
```diff
@@ -0,0 +1,80 @@
...
+| CentOS Stream | Jan 1st, 2022 | N/A | N/A | N/A | N/A |
```
alexjfisher, line 42: I *think* there is an EOL date as there will be a CentOS 8 Stream and later a CentOS 9 Stream.

_docs/software_support_lifecycle.md
```diff
@@ -0,0 +1,80 @@
...
+        1. Major
+        2. Minor
+        3. Patch
+        4. Examples: 
+            1. `5.1.1` is the latest version and supported.
+            2. `5.1.0` is not supported.
+            3. `4.9.9` is not supported.
+    2. Exceptions:
+        1. The maintainer may make exceptions to the above rule as long as it does not create undue expectations for the community.
+2. For all other documenation on Archived Modules see [Deprecated and Archived Modules]({% doc_url deprecated_and_archived_modules %})
```
alexjfisher, line 80: spelling of `documentation`

_docs/software_support_lifecycle.md
alexjfisher: I think it's a 'wait and see' for now as to how we test RHEL 8 when CentOS 8 disappears.  Maybe we just have to use CentOS 8.3 images even when RHEL is on eg 8.8???

_docs/software_support_lifecycle.md
dhollinger: I believe the CentOS itself only supports the mainline CentOS Stream. CentOS Stream 8 and Stream 9 are for RHEL Dev purposes.

dhollinger:
```
Each major release will have a branch, similar to how CentOS Linux is currently structured; however, CentOS Stream is designed to focus on RHEL development, so only the latest Stream will have the marketing focus of the CentOS Project.
```

https://centos.org/distro-faq/

_docs/software_support_lifecycle.md
dhollinger: According to CentOS' FAQ, we need to be testing for RHEL on RHEL images going forward :shrug: 

@ekohl mentioned that there seems to be some confusion around this though.

_docs/software_support_lifecycle.md
```diff
@@ -0,0 +1,80 @@
...
+| OpenSUSE Leap 15.1 | May 22nd, 2019 | Jan 31st, 2021 | Dec 31st, 2020 | Jan 31st, 2021 | Jan 31st, 2022 |
+| OpenSUSE Leap 15.2 | July 2nd, 2020 | Dec 31st, 2021 | Sept 30th, 2021 | Dec 31st, 2021 | Dec 31st, 2022 |
+
+#### Puppet Versions
+
+Puppet support will fall in-line with the Puppet Open Source project's lifecycle.
+
+1. A Puppet version that has been labelled as End of Life by the Puppet Open Source project, will no longer be supported by Vox Pupuli.
+2. A deprecation PR will be created no later than 6 months before End of Life.
+3. All tests and data in metadata.json will be removed no later than the End of Life date for that version.
```
igalic, line 60: i don't remember this specific outcome of our discussions
if this is happening, then we should branch a 5.x and allow the community to create PRs against that

igalic (commented):
🙋🏻️

_docs/software_support_lifecycle.md
```diff
@@ -0,0 +1,80 @@
...
+To account for this, we no longer support any software or operating system after the date that the maintainers have marked said software or operating system as End of Life. However, this does not mean we will immediately remove code that allows for the module to run on End of Life'd software or operating systems, but the community will no longer be expected to provide support to users for that software.
+
+There are exceptions and the full policy is detailed in the next section.
+
+### Support Lifecycle
+
+#### Operating Systems
+An Operating System is no longer considered supported once it reaches an End of Life date determined by the upstream maintainer* of the operating system. To ensure our users have enough time to adjust, we will take the following steps to ensure as seamless of a transition as possible:
+
+1. A deprecation PR will be created no later that 3 months before an Operating System's End of Life date.
```
ghoneycutt, line 23: Do you mean "no earlier than" ?

_docs/software_support_lifecycle.md
ghoneycutt: To avoid confusion it would be beneficial to outline the specifics so if people want to do it, there is a pattern. Like creating a branch named `puppet5`.

_docs/software_support_lifecycle.md
ekohl: I think a PR 3 months in advance is good. However, is the intent also to merge it or is it to open a discussion? For a discussion, an issue may be good.

Perhaps this is also something we can script, using metadata.json. They should also have a label so it's easy to find them.

_docs/software_support_lifecycle.md
ekohl: For those unaware, I have sent an email about this with the PMC on CC. There is nothing secret in [the initial email](https://gist.github.com/ekohl/6418d1435b9f12a394256bb0367c6a0b). To set the expectations right, I think the people responsible for that alias are currently busy but they are working on a solution exactly for the problem we're having: CI to test integration for community projects. While I work for Red Hat, this is not an area I work on and can only advocate for it. Luckily, our use case is not special in any way (loads of projects can/should do CI like we do) and one that was explicitly mentioned to be a focus point.

_docs/software_support_lifecycle.md
ekohl: It is my understanding that while RHEL 8 is its full support cycle, there will be a CentOS 8 stream. During RHEL's full support cycle, new features and hardware enablement are being implemented and those should first land in CentOS 8 stream. It is exactly the purpose of stream to allow anyone a preview of that. The goal here is Continuous Delivery, similar to how modern development should have an always production ready git master branch. There should be CI to cover this, but of course there will be bugs and regressions since it's still software.

After the full support cycle ends, it enters Maintenance Support. In this phase, RHEL should only receive security features and urgent bug fixes. As such, there's a lot less need for integration testing for upcoming releases (essentially what stream allows).

There is a period (roughly 2 years) where development of RHEL 8 and 9 overlap. During this period there should be 2 streams in parallel, but I think marketing focus implies they'll direct users to CentOS Stream 9 rather than 8.

This is based on https://access.redhat.com/support/policy/updates/errata/

_docs/software_support_lifecycle.md
```diff
@@ -0,0 +1,80 @@
...
+#### Puppet Versions
+
+Puppet support will fall in-line with the Puppet Open Source project's lifecycle.
+
+1. A Puppet version that has been labelled as End of Life by the Puppet Open Source project, will no longer be supported by Vox Pupuli.
+2. A deprecation PR will be created no later than 6 months before End of Life.
+3. All tests and data in metadata.json will be removed no later than the End of Life date for that version.
+4. Code that works on the End of Life versions will not be removed unless it bars the progress of the module.
+5. Pull Requests may still be accepted, as long as it doesn't hinder forward progress of the module or add End of Life-specific code.
+6. Issues will no longer be accepted for the End of Life version.
```
ekohl, line 63: Perhaps it should be phrased as something that should affect supported version. It's fine if they open an issue with "I saw this on 5.5, but later reproduced on 6.10". We should not discourage that.
```suggestion
6. Issues are only valid if they affect supported versions.
```

_docs/software_support_lifecycle.md
kenyon: A Red Hat Developer subscription for the Vox Pupuli organization would almost work, if not for the "single-user" limitations. https://developers.redhat.com/articles/faqs-no-cost-red-hat-enterprise-linux

ekohl: I have been told they want to fix this. Since it's a license, legal is involved. That means we need to be patient. The holiday season probably doesn't help either.

_docs/software_support_lifecycle.md
```diff
@@ -0,0 +1,80 @@
...
+As we operate entirely on the graces of our community members being willing to voluntarily contribute to our modules, we also need to be respective of their time and resources and not add undue stress in order to support software that upstream creators no longer support.
+
+To account for this, we no longer support any software or operating system after the date that the maintainers have marked said software or operating system as End of Life. However, this does not mean we will immediately remove code that allows for the module to run on End of Life'd software or operating systems, but the community will no longer be expected to provide support to users for that software.
+
+There are exceptions and the full policy is detailed in the next section.
+
+### Support Lifecycle
+
+#### Operating Systems
+An Operating System is no longer considered supported once it reaches an End of Life date determined by the upstream maintainer* of the operating system. To ensure our users have enough time to adjust, we will take the following steps to ensure as seamless of a transition as possible:
```
bastelfreak, line 21: there are some edge cases. for example debian has still "support" for debian 8, but that's very very limited. In the past we considered that as not-supported. Same applies for RedHat 5/6 where you can only get support if you buy it from redhat. I also consider those distros as not supported anymore

_docs/software_support_lifecycle.md
```diff
@@ -0,0 +1,80 @@
...
+
+1. A deprecation PR will be created no later that 3 months before an Operating System's End of Life date.
+2. Removal of the End of Life version from the metadata.json and tests to be completed by the End of Life date.
+3. Issues relating to the End of Life operating system will no longer be addressed after the End of Life date.
+4. Code allowing a module to be run on the End of Life operating system will be left in the module until one of the following milestones are reached after which time the code will be removed:
+    1. Maintaining the code prevents the module from moving forward.
+    2. 12 Months after the End of Life date.
```
bastelfreak, line 28: I don't think that this is realistic. It already requires a bunch of time to go through modules and remove an OS from metadata.json, review the tests, get that merged, make a new release. I won't recheck the module a year later to drop dead code. I still believe code related to an EOL OS should be removed when we purge the OS from the metadata.json

ekohl: Agreed. Dead code is bad. It's confusing to new contributors. Having code implies support, but how do you know it actually works? I'm also in favor of a hard cut. Dropping it means dropping it. No grace period where you have Schödinger's support.

_docs/software_support_lifecycle.md
```diff
@@ -0,0 +1,80 @@
...
+title: Software Support Lifecycle
+date: 2020-12-09
+summary: Vox Pupuli policy on supporting upstream software
+---
+
+## Software Support Lifecycle
+
+As a part of Vox Pupuli's mission to provide a home for modules that still have a userbase and need development, we need to define a clear support strategy for what operating systems, dependencies, and upstream software we will support and for how long those will be supported.
+
+As we operate entirely on the graces of our community members being willing to voluntarily contribute to our modules, we also need to be respective of their time and resources and not add undue stress in order to support software that upstream creators no longer support.
```
mmerfort, line 12: Should be “respectful” instead of “respective”.

_docs/software_support_lifecycle.md
```diff
@@ -0,0 +1,80 @@
...
+
+To account for this, we no longer support any software or operating system after the date that the maintainers have marked said software or operating system as End of Life. However, this does not mean we will immediately remove code that allows for the module to run on End of Life'd software or operating systems, but the community will no longer be expected to provide support to users for that software.
```
mmerfort, line 14: Maybe define End of Life as EOL here and use EOL below instead.

_docs/software_support_lifecycle.md
mmerfort: I agree with @bastelfreak that removing code for unsupported functionality is a good idea. People can just use the version which still supports their old OS or software version.

If we don't tightly couple removing unsupported features to deprecations, it might actually make it more difficult for contributors since they now have to work around this code. For these contributors, it's not going to always be obvious if code is part of the deprecated feature set.

_docs/software_support_lifecycle.md
dhollinger: There can be some leeway for `puppet5` due to the timing, but I am strongly against putting the pressure on an already over-strained Vox Pupuli community to support EoL software in the future.. 
It'd be different if we were a paid team offering Enterprise support, but we are not.

_docs/software_support_lifecycle.md
dhollinger: I said "No later than", so that ample time is given to the user. We had a long discussion on the PMC call about it recently and the idea was that, like any Enterprise software, if a user wants support post-EoL date, they have to pay a vendor or support it themselves.

Vox Pupuli doesn't have the manpower to support EoL OSes and Software. It'd be different if every EoL date was like CentOS 8, but in reality, most of those dates are out there a year or more ahead of time, so I'm strongly against supporting past an EOL date.

However, "End of Support" does not mean we should break it intentionally either. We leave functionality in there, if possible, until 1 year after the EOL date. The exception being that if a currently supported/stable version needs something that breaks the EoL version, then the supported versions take precedence.

_docs/software_support_lifecycle.md
dhollinger: I think we do need to decide ahead of time which versions we will support.

For Debian, do we want to support until EOL or until End of LTS. This one is confusing as other LTS operating systems, like RHEL and Ubuntu, have EOL and end of LTS as the same thing and post LTS support as "Extended Support".

I think for RHEL, I'd like to end our support on the EOL date, not the End of Extended support. That only affects paying RHEL customers and we aren't in the "business" of supporting Enterprises. We build for any user and as such our focus needs to be on what's best for the community as a whole, not the Enterprise users only.

_docs/software_support_lifecycle.md
dhollinger: If this is the prevailing desire and there isn't anyone who wants to push back, then I'll add this into the document tomorrow.

_docs/software_support_lifecycle.md
dhollinger: Oh my goodness, that is horrible and entirely on me.

_docs/software_support_lifecycle.md
dhollinger: Updated with date change and links

_docs/software_support_lifecycle.md
ekohl: Normally I'm inclined to agree. I certainly agree we're already over-strained. However, removing Puppet 5 code is typically more work than keeping it in. Except for the systemd daemon-reload fix in Puppet 6.1 I'm not aware of things that are really different between 5 and 6. That means it's typically only an entry in metadata.json that's different (which results in a different test matrix).

_docs/software_support_lifecycle.md
ekohl: I'm opposed to having dead code in modules, but it's not black and white. I'll give some examples that I can think of. In the following examples OS major version 6 is EOL, 7 isn't.

```puppet
# No need to change 6 to 7, that's not really a difference and only creates noise in the git log
if versioncmp($facts['os']['release']['major'], '6') > 0 {}

# This is code that's not run on any supported OS which means it's not covered by tests
if $facts['os']['release']['major'] == '6' {}
```

So while it's annoying, I do think the rule of thumb should be that we can test all code.

I also think that individual contributors should be able to support it if they provide the manpower to keep it supported. Let's say I need a module on an EOL OS, then if I make sure it's all working, that should be OK. This should be documented somewhere (README?). It is also an explicit process. During the period before it goes EOL, it must be noted by a project member that they are interested in keeping it up. Given our current project composition, this may not be used that much. I know I generally have no interest in it since I keep my distros supported.

dhollinger: I would also add the caveat that keeping up code to work with an EOL OS would only be allowed as long as it doesn't prevent the project from moving forward to add features or support current OSes.

ekohl: Btw, I think https://github.com/puppetlabs/puppetlabs-apache is a good example of the cruft that can build up. It still has conditionals for Ubuntu 10.04.

_docs/software_support_lifecycle.md
```diff
@@ -0,0 +1,86 @@
...
+
+##### Support Table:
+
+| OS Name | Support Begins | EOL | Notification PR | End of Support | Removal of Code |
+| ------- | -------------- | ----------- | --------------- | -------------- | --------------- |
+| CentOS/RHEL 6 | 2011/07/10 | 2020/11/30 | N/A | 2020/11/30 | 2021/11/30 |
+| CentOS/RHEL 7 | 2014/07/07 | 2024/06/30 | 2024/03/31 | 2024/06/30 | 2025/06/30 |
+| CentOS 8 | 2019/09/24 | 2021/12/31 | 2021/09/30 | 2021/12/31 | 2022/12/31 |
+| RHEL 8 |  2019/05/01 | 2029/05/31 | 2029/02/28 | 2029/05/31 | 2030/05/31 |
+| CentOS Stream 8 | 2022/01/01 | 2029/05/31 | 2029/02/28 | 2029/05/31 | 2030/05/31 |
```
tmannerm, line 43: CentOS Stream 8 upstream EOL is in May 31, 2024 according to https://centos.org/distro-faq/#q2-what-about-the-other-releases-of-centos-linux that states it's supported until "full RHEL support phase" which is defined in https://access.redhat.com/support/policy/updates/errata/#Life_Cycle_Dates and specifically doesn't include the 5 year Maintenance Phase RHEL itself gets. There will be CentOS Stream 9 with 2 year overlap released before that according to Question 6 of the FAQ, though.

_docs/software_support_lifecycle.md
bastelfreak: There weren't any new people voting for keeping the code a year, so I guess it's safe to say that the majority is fine with dropping the code in the moment where we drop the OS from metadata.json?

towo: Same context @bastelfreak is asking from: does everybody consider PR CI being red as «maintaining the code prevents the module from moving forward»? Technically it's not actually stopping it, but while one might be inclined to check if it's okay for a certain set of a module's CI checks to be failing because they're from an EOL OS, it's a mental load issue.

Sure, it's possible to handle manually parsing if a CI is allowed to be red, but I'd argue that as soon as an EOL OS bumps the CI into the red, it's ripe for removal; at the very least from metadata.json if *only* the OS-specific CI is an issue, and just wholesale otherwise.

ekohl:
So how do we break the stalemate here? It appears some people are in favor of keeping dead code while others (myself included) are against. Would a vote help?

binford2k:
Let's commence a vote then!

👍 Keep dead code until one of the listed milestones is reached
👎 Kill dead code immediately and require users to pin to old versions at their own peril

ekohl:
The votes look fairly conclusive to me but let's give this an explicit deadline. Is next week (March 8th) ok for everyone?

ekohl:
Today is the 8th. I don't think the votes changed and we can conclude that with 9 votes `Kill dead code immediately and require users to pin to old versions at their own peril` won. There were 2 votes for `Keep dead code until one of the listed milestones is reached`. It should be noted that @binford2k voted for both, likely so both buttons showed up.

Thanks everyone who voted.

@dhollinger can you update the PR or anyone else wants to take a stab?

binford2k:
@ekohl yes, that was my intention.

dhollinger:
I will take a look shortly and update the PR accordingly

_docs/software_support_lifecycle.md
```diff
@@ -0,0 +1,84 @@
...
+
+As a part of Vox Pupuli's mission to provide a home for modules that still have a userbase and need development, we need to define a clear support strategy for what operating systems, dependencies, and upstream software we will support and for how long those will be supported.
+
+As we operate entirely on the graces of our community members being willing to voluntarily contribute to our modules, we also need to be respectful of their time and resources and not add undue stress in order to support software that upstream creators no longer support.
+
+To account for this, we no longer support any software or operating system after the date that the maintainers have marked said software or operating system as End of Life, hereafter referred to as "EOL". However, this does not mean we will immediately remove code that allows for the module to run on EOL'd software or operating systems, but the community will no longer be expected to provide support to users for that software. 
+
+If a user still needs to use the EOL operating system, then they must either:
+* Pin their module version to a version compatible with their operating system.
+* Fork the module and maintain it internally.
```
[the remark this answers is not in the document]
bastelfreak: yeah. I think as long as the person is willing to maintain support and it's tested and it doesn't block further development it's fine. but that's also an edge case.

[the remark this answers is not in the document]
binford2k: Should modules with extended support policies have a clearly marked exception to the standard case?

_docs/software_support_lifecycle.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Support Lifecycle
+
+#### Operating Systems
+An Operating System is no longer considered supported once it reaches an EOL date determined by the upstream maintainer* of the operating system. To ensure our users have enough time to adjust, we will take the following steps to ensure as seamless of a transition as possible:
+
+1. A deprecation PR will be created no later that 3 months before an Operating System's EOL date.
+2. Removal of the EOL version from the metadata.json and tests to be completed by the EOL date.
+3. Issues relating to the EOL operating system will no longer be addressed after the EOL date.
+4. While a module will not have general code replace, all code specific to the EOL operating system will be removed by the EOL date, or shortly thereafter.
```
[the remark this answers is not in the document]
bastelfreak: cleaned it up

kenyon (approved):

bastelfreak:
I will merge this now as-is. In my opinion it reflects the current state of the discussion. The PR has quite a history of comments and more comments make it even harder to read. If anything doesn't agree with the post please raise an issue or open a PR to fix it.

PR #234, opened 2021-01-25
Add empty_trailing_lines to puppet-lint plugin list

ghoneycutt:


ghoneycutt:
This plugin has been used for awhile with the popular Sensu module (https://github.com/sensu/sensu-puppet/blob/master/Gemfile#L29). It also has the ability to fix and has spec tests.

bastelfreak (approved):

PR #240, opened 2021-06-15
document that we delete merged branches

bastelfreak:
We enable `Automatically delete head branches` since a long time so we
should also document it.


smortex (approved):

kenyon:
Travis seems stuck 😭

PR #242, opened 2021-06-18
dont advise people to release in the old namespace

bastelfreak:
we don't do that since a long time anymore, because the forge finally
has a deprecation feature. that's used instead and contains a notice.


ekohl:
(commit message) Isn't it `advise` in this case?

llowder (approved):

PR #253, opened 2021-10-29
Clarify handling for optional literal parameters

bastelfreak:
This mentions that parameters should default to undef and not empty string. We check for that in reviews since ages, I thought we already had it in the document. But that wasn't the case.


_docs/reviewing_pr.md
```diff
@@ -11,7 +11,7 @@ one of our modules:
 * Does the email address used in the commits match the github email address? (This will let github display the contributor's avatar next to the commit)
 * Is this a bugfix, modulesync, breaking change, enhancement, docs update? Label it with `bug`, `modulesync`, `backwards-incompatible`, `enhancement`, `docs`
 * Are updates to the README.md needed but missing? Label it with `needs-docs`
-* Has the file documented params or examples in the header? This needs to be updated as well
+* Has the file documented parameter or examples in the header? This needs to be updated as well
```
jstraw, line 14: parameters should still be plural here to match examples
```suggestion
* Has the updated file(s) documented all of its parameters or and examples in the puppet-strings header? This needs to be updated as well.
```

_docs/reviewing_pr.md
```diff
@@ -22,8 +22,9 @@ one of our modules:
...
 * Are hiera yaml files added for data-in-modules? Ensure that the data is compatible with [hiera 5](https://puppet.com/docs/puppet/5.3/hiera_migrate.html#use-cases-for-upgrading-to-hiera-5). Static data that is equal across every supported operating system must stay in the init.pp, it shouldn't be moved to a common.yaml due to [puppet-strings issue #250](https://github.com/puppetlabs/puppet-strings/issues/250).
-* Are there new params with datatype Hash or Array? If possible, they should default to empty Hash/Array instead of undef. You can also enforce the datastructure like Array[String[1]]
-* Are there new params with datatype Boolean? The default value is a tricky decision which needs careful reviewing. Sometimes a True/False is the better approach, sometimes undef
+* Are there new parameters with datatype Hash or Array? If possible, they should default to empty `Hash` or `Array` instead of `undef`. You can also enforce the datastructure like `Array[String[1]]` (that's recommended if possible)
```
jstraw, line 25: Do we want to recommend fully realized data structures if it's obvious what they should be, or type aliases for more complex things?  so Hash[String, Hash[String, Array[Any]]] for most create_resources collectors

jstraw (changes requested):

_docs/reviewing_pr.md
bastelfreak: good catch. I added:
`* Is the new datatype quite complex (like `Hash[String[1], Hash[String[1], Array[Any]]]`? Create a custom datatype for it.`

jstraw (approved):

PR #254, opened 2021-11-02
Document params_empty_string/deprecate empty_string

bastelfreak:


TuningYourCode (approved):

jstraw (approved):
It might be worth describing why deprecation occurred, but I know it's documented in the other repo.

PR #260, opened 2022-07-18
Document PQL queries

bastelfreak:
A few times we had discussions on IRC and slack about a page where people can document various PQL queries. I think this is a good start.

### DIFF
--- a/_docs/pql_queries.md
+++ b/_docs/pql_queries.md
@@ -23,6 +23,8 @@ puppet resource package puppetdb_cli ensure=present provider=puppet_gem
 
 And Afterwards run the query with `/opt/puppetlabs/puppet/bin/puppet-query $query`. For PE installations you can use `puppet query`.
 
+On the PuppetServer side, installing the `puppetdb-termini` package allow to use the `puppetdb_query()` function in a manifest to generate portions of catalog based on result of queries, e.g. find all Virtual-Hosts on all machines and add the relevant configuration to a monitoring system.
+
 ## Contribute?!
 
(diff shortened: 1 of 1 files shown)


_docs/pql_queries.md
```diff
@@ -0,0 +1,73 @@
...
+
+## How?
+
+All queries here are documented in the [string-based query style](https://puppet.com/docs/puppetdb/7/api/query/v4/pql.html), not the [AST](https://puppet.com/docs/puppetdb/7/api/query/v4/ast.html) style. You need the [Puppet Client tools](https://puppet.com/docs/puppetdb/7/pdb_client_tools.html) (for Open Source or PE) to get the `puppet query` face. To install the client tools on Open Source, you can do:
+
+```
+puppet resource package puppetdb_cli ensure=present provider=puppet_gem
+```
+
+And Afterwards run the query with `/opt/puppetlabs/puppet/bin/puppet-query $query`. For PE installations you can use `puppet query`.
```
smortex, line 24:
```suggestion
And Afterwards run the query with `/opt/puppetlabs/puppet/bin/puppet-query $query`. For PE installations you can use `puppet query`.

On the PuppetServer side, installing the `puppetdb-termini` package allow to use the `puppetdb_query()` function in a manifest to generate portions of catalog based on result of queries, e.g. find all Virtual-Hosts on all machines and add the relevant configuration to a monitoring system.
```

smortex (commented):
Maybe we can add a note about `puppetdb_query()`?

baurmatt (approved):

PR #261, opened 2022-07-21
Additional queries and general endpoint info

hbrown-uiowa:
I have notes to myself about which endpoints are available which I find useful. Feel free to discard any tip that isn't useful. I also wasn't sure on ordering of the extra info about endpoints and fields.


bastelfreak:
awesome, thanks!

PR #266, opened 2022-07-25
document more PQL queries

bastelfreak:


jcpunk (approved):

_docs/pql_queries.md
```diff
@@ -95,6 +95,20 @@ puppet query 'reports[certname,transaction_uuid,receive_time] { cached_catalog_s
 puppet query 'inventory[certname] { facts.os.name = "windows" }'
 ```
 
+### Get a list of nodes with to specific facts
```
alexjfisher, line 98:
```suggestion
### Get a list of nodes with two specific facts
```

bastelfreak: thanks, will raise another PR

PR #269, opened 2022-09-28
Document gem release workflow

bastelfreak:


_docs/releasing_gem.md
```diff
@@ -0,0 +1,61 @@
...
+This pull request updates the changelog and bumps the version number to the
+target version. The version is set in the gemspec file in the root of the git
+repository. Sometimes it's a variable coming from something like
+`lib/$gem/version.rb`. An example is [beaker](https://github.com/voxpupuli/beaker/blob/f60ac9413f9c7976a6645ef9e1dd2afbcc6542de/beaker.gemspec#L8),
+([version.rb](https://github.com/voxpupuli/beaker/blob/master/lib/beaker/version.rb#L3)).
+
+Now you can install the changelog generator:
+
+```bash
+bundle install --jobs $(nproc) --path .vendor/ --with release
```
ekohl, line 22: I think passing `--path` and `--with` to `bundle install` is deprecated. Also, for parallelism I'd suggest referring to some common documentation. `export BUNDLE_JOBS="$(nproc)"` in my `~/.bashrc`.

_docs/releasing_gem.md
```diff
@@ -0,0 +1,61 @@
...
+backwards-incompatible nor only bug fixes, make a minor release. Check the
+generated diff for the CHANGELOG.md. If your chosen release version doesn't
+match the generated changelog, update gemspec and run the changelog task again.
+
+Get community feedback on the release pr, label it with skip-changelog, get it merged.
+
+Checkout an updated copy of master
+
+```bash
+git checkout master; git fetch origin; git pull origin master
```
ekohl, line 50: `git pull` is equal to `git fetch && git merge`.
```suggestion
git checkout master; git pull origin master
```

_docs/releasing_gem.md
bastelfreak: yes it's deprecated. I need to check for the new syntax, I cannot keep that in my head :D

bastelfreak: I updated this

_docs/releasing_gem.md
```diff
@@ -0,0 +1,63 @@
...
+repository. Sometimes it's a variable coming from something like
+`lib/$gem/version.rb`. An example is [beaker](https://github.com/voxpupuli/beaker/blob/f60ac9413f9c7976a6645ef9e1dd2afbcc6542de/beaker.gemspec#L8),
+([version.rb](https://github.com/voxpupuli/beaker/blob/master/lib/beaker/version.rb#L3)).
+
+Now you can install the changelog generator:
+
+```bash
+bundle config set --local path '.vendor/'
+bundle config set --local with 'release'
+bundle install --with release
```
smortex, line 24:
```suggestion
bundle install
```

We set this config on the line above

_docs/releasing_gem.md
```diff
@@ -0,0 +1,63 @@
...
+
+The changelog generator checks for certain labels on closed issues and PRs since
+the last release and groups them together. If the changes were neither
+backwards-incompatible nor only bug fixes, make a minor release. Check the
+generated diff for the CHANGELOG.md. If your chosen release version doesn't
+match the generated changelog, update gemspec and run the changelog task again.
+
+Get community feedback on the release pr, label it with skip-changelog, get it merged.
+
+Checkout an updated copy of master
```
smortex, line 49: (bikeshelding) We can separate the two parts with sections (not sure about the number of `#`) and add something like this above.

```
## Part 2: Actually create the release
```

These creates anchors, so in the header we can say it is a two step process, list them and link them to the actual sections.

_docs/releasing_gem.md
```diff
@@ -0,0 +1,63 @@
...
+
+```bash
+git checkout master; git pull origin master
+```
+
+check with `git tag -l` if the git tags are prefixed with a v or not. This varies
```
smortex, line 55: (bikeshelding) Capitalization of sentence.

bastelfreak: fixed

_docs/releasing_gem.md
bastelfreak: fixed

_docs/releasing_gem.md
```diff
@@ -0,0 +1,65 @@
...
+git checkout master; git pull origin master
+```
+
+Check with `git tag -l` if the git tags are prefixed with a v or not. This varies
+by project (we often adopt gems from other people and don't want to change the
+versioning scheme in a project).
+
+Create a new git tag with the new version: `git tag -s $version`
+
+Push the git tag: `git push origin $version`
```
ekohl, line 63: I always prefer `git push --follow-tags`. I think this actually only pushes the tag.

_docs/releasing_gem.md
```diff
@@ -0,0 +1,65 @@
...
+
+```bash
+git checkout master; git pull origin master
+```
+
+Check with `git tag -l` if the git tags are prefixed with a v or not. This varies
+by project (we often adopt gems from other people and don't want to change the
+versioning scheme in a project).
+
+Create a new git tag with the new version: `git tag -s $version`
```
ekohl, line 61: Why do we manually tag here? We have a rake task for it. https://voxpupuli.org/docs/releasing_version/ says we should use `rake release`.

ekohl (commented):
Late to the party, but this was caught in my backlog :(

_docs/releasing_gem.md
ekohl: Ah, my bad. This is gems, not modules.

PR #281, opened 2023-02-26
pql_queries: add languages to code blocks

kenyon:
And a couple of other minor [markdownlint](https://github.com/DavidAnson/markdownlint) fixes.

### DIFF
--- a/_docs/pql_queries.md
+++ b/_docs/pql_queries.md
@@ -20,7 +20,7 @@ The syntax isn't alway very intuitive. That's why this page exists. It serves as
 
 All queries here are documented in the [string-based query style](https://puppet.com/docs/puppetdb/7/api/query/v4/pql.html), not the [AST](https://puppet.com/docs/puppetdb/7/api/query/v4/ast.html) style. You need the [Puppet Client tools](https://puppet.com/docs/puppetdb/7/pdb_client_tools.html) (for Open Source or PE) to get the `puppet query` face. To install the client tools on Open Source, you can do:
 
-```
+```shell
 puppet resource package puppetdb_cli ensure=present provider=puppet_gem
 ```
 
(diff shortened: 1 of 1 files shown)


_docs/pql_queries.md
```diff
@@ -36,13 +36,13 @@ You can submit your own Queries by editing [voxpupuli.github.io/_docs/pql_querie
...
 puppet query 'nodes[certname] {}'
 ```
 
 Result:
 
-```
+```json
 [
   {
     "certname": "puppetserver.bastelfreak.org"
```
ekohl, line 48: Unrelated, but probably best to use `example.org` instead.

_docs/pql_queries.md
```diff
@@ -20,7 +20,7 @@ The syntax isn't alway very intuitive. That's why this page exists. It serves as
 
 All queries here are documented in the [string-based query style](https://puppet.com/docs/puppetdb/7/api/query/v4/pql.html), not the [AST](https://puppet.com/docs/puppetdb/7/api/query/v4/ast.html) style. You need the [Puppet Client tools](https://puppet.com/docs/puppetdb/7/pdb_client_tools.html) (for Open Source or PE) to get the `puppet query` face. To install the client tools on Open Source, you can do:
 
-```
+```console
```
ekohl, line 23: This doesn't look like `console`, which is a command prefixed by either `#` (for root) or `$` (for non-root). Then unprefixed lines are the command output. If anything, this is `shell`.

ekohl (changes requested):

_docs/pql_queries.md
kenyon: Yeah, it was a toss up since they aren't colorized either way, mostly just like to eliminate warnings from vscode. I'll go with `shell`.

_docs/pql_queries.md
kenyon: Changed to `example.org`.

_docs/pql_queries.md
kenyon: Ah actually `shell` does color the quoted parts. 👍

_docs/pql_queries.md
```diff
@@ -206,11 +206,12 @@ Result:
...
-The available endpoints is a function of which version of puppetdb you are going against. The current list is available at https://puppet.com/docs/puppetdb/7/api/query/v4/entities.html
+
+The available endpoints is a function of which version of puppetdb you are going against. The current list is available at <https://puppet.com/docs/puppetdb/7/api/query/v4/entities.html>.
 
 If you don't feel like looking up which fields are available for a given endpoint, use a bogus field name and the `puppet query` tool will return the valid list.
 
-```
-puppet-query 'resources[foo] {}'
+```console
+$ puppet-query 'resources[foo] {}'
```
bastelfreak, line 215:
```suggestion
$ puppet query 'resources[foo] {}'
```

bastelfreak (approved):

PR #288, opened 2023-04-19
Clarify module release documentation

natemccurdy:
This updates the module release documentation to it flows as a series of steps.
Hopefully this makes it easier to follow along for people not familiar with the process.

### DIFF
--- a/_docs/releasing_version.md
+++ b/_docs/releasing_version.md
@@ -6,68 +6,99 @@ summary: How to perform a complete version release, including modulesync and pub
 ---
 
 Creating a release is a two step process:
-1. Prepare the release — setup everything so that peer review can happen and when everything is ready…
+1. Prepare the release
 2. Do the actual release
 
 ## Preparing a release
 
-Run modulesync to ensure the dotfiles are up-to-date.
+Before preparing a release, it may help to run modulesync to ensure the
+dotfiles are up-to-date. If modulesync has been run recently by someone else,
+you may not have to do it. See https://github.com/voxpupuli/modulesync for more
+info on using modulesync.
 
-Create a 'release pr'. This pull request updates the changelog and bumps the
-version number to the target version, removing all release candidate
-identifiers, i.e. from `0.10.7-rc0` to `0.10.7`. Here's an example:
+### 1. Create a new branch for a "release pull request"
+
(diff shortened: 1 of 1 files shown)


_docs/releasing_version.md
```diff
@@ -6,69 +6,100 @@ summary: How to perform a complete version release, including modulesync and pub
...
+
+> If necessary, run `bundle install` before continuing.
+> ```bash
+> bundle install --path .vendor/ --without system_tests development
+> ```
+>
+> And in case you installed the gems before:
+> ```bash
+> bundle install --path .vendor/ --without system_tests development; bundle update; bundle clean
+> ```
```
kenyon, line 57: I wonder if this should be simplified to just always doing the install/update/clean command, so that you don't have to know the answer to when it would be necessary to run `bundle install` before continuing.

kenyon (approved):

_docs/releasing_version.md
```diff
@@ -6,69 +6,100 @@ summary: How to perform a complete version release, including modulesync and pub
...
+1. Prepare the release
 2. Do the actual release
 
 ## Preparing a release
 
-Run modulesync to ensure the dotfiles are up-to-date.
+Before preparing a release, it may help to run modulesync to ensure the
+dotfiles are up-to-date. If modulesync has been run recently by someone else,
+you may not have to do it. See https://github.com/voxpupuli/modulesync for more
+info on using modulesync.
```
bastelfreak, line 17: do you think this is fine or should we enhance it with a sentence so people can check if it's actually up2date? each of our modules has the `.msync.yml` file with the version:
```
$ cat .msync.yml
---
# Managed by modulesync - DO NOT EDIT
# https://voxpupuli.org/docs/updating-files-managed-with-modulesync/

modulesync_config_version: '5.5.0'
```
Which maps to the tag at: https://github.com/voxpupuli/modulesync_config/tags

also maybe it's better to link to https://github.com/voxpupuli/modulesync_config/ and not https://github.com/voxpupuli/modulesync ?

_docs/releasing_version.md
bastelfreak: that's the reason why I initially provided it as a long line. just copy paste it and it will set everything up.

PR #306, opened 2023-07-20
Attempt to improve the website

smortex:
Yeah, I am not a web developer!  We have inconsistencies and blatant UI issues that this PR attempts to solve.

We depend on outdated third party "things" (e.g. Material Design for Bootstrap) that should probably be updated so that we can follow the current documentation, but that would be more effort that what I can offer at the moment.

| Before | After |
|---|---|
| Home page, before | Home page, after |
| Coc page, before | Coc page, after |
| Documentation, before | Documentation, after |
| Sponsors, before | Sponsors, after |

<details>
<summary>Unmentionable request</summary>

Meme image of I can has cheezburger cat with "I can haz a webdeveloper badge?"

</details>


_config.yml
```diff
@@ -21,6 +21,8 @@ kramdown:
 collections:
   docs:
     output: true
+  sponsors:
+    output: false
```
smortex, line 25: With a collection we can just manage a bunch of small files in the `_sponsors` directory to add/remove sponsors.  For now, they are sorted by name, but if we want to sort them further this seems to be supported:
https://jekyllrb.com/docs/collections/#custom-sorting-of-documents

static/images/Makefile
smortex, line 1: In order to avoid misaligments of the sponsors logos, the easiest way is to have all of them use the same aspect ratio.  For this purpose I put a Makefile in the hope that when we want to add more logos, we will be able to quickly find how to generate them without too much effort.

static/images/Makefile
```diff
@@ -0,0 +1,31 @@
...
+	convert "logo-Enterprise-modules SVG.png" -background transparent -resize ${SPONSOR_LOGO} -gravity center -extent ${SPONSOR_CANVAS}  sponsor-enterprise-modules.png
+
+hetzner_cloud_logo.png: hetzner_cloud_logo.svg
+	inkscape -o hetzner_cloud_logo.png hetzner_cloud_logo.svg
+
+babiel_logo_RGB.png: babiel_logo_RGB.svg
+	inkscape -o babiel_logo_RGB.png babiel_logo_RGB.svg
+
+logo-Enterprise-modules\ SVG.png: logo-Enterprise-modules\ SVG.svg
+	inkscape -o "logo-Enterprise-modules SVG.png" "logo-Enterprise-modules SVG.svg"
```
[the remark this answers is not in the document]
smortex: Nice! This was a quick hack and I am not sure if we want to keep it that way or not, but nevertheless this is great :-)

_includes/nav.html
```diff
@@ -20,12 +20,13 @@
       {% endfor %}
     </ul>
     <ul class="nav navbar-nav navbar-right">
-      <li><a href="https://voxpupuli.org/privacy-policy"> Privacy Policy</a></li>
+      <li><a href="/privacy-policy"> Privacy Policy</a></li>
```
ekohl, line 23: Should this also use the site.url or site.baseurl?

smortex: We can, but this trend to be cumbersome.  I wondered if trimming `{{ site.url }}{{ site.baseurl }}` everywhere (and also the `{{ site.url }}` where `{{ site.baseurl }}` was forgotten) would make more sense, and then saw that Jekyll 4 has a `link` function that raise an error when we link to something that does not exist: https://jekyllrb.com/docs/liquid/tags/#link

Unfortunately we are running an older version of Jekyll ATM (github-pages require jekyll 3.9.3), but it would be a nice move in the future, so I will put '{{ site.url }}{{ site.baseurl }}' everywhere so that we can change it when we can.

smortex: Done

bastelfreak (approved):

PR #310, opened 2023-08-08
update and clarify what has to be done where

rwaffen:
i stumbled upon this while doing my first releases and thought it could be good to know what you do where.
also i generalized when to do a major bump and remove the ruby1.8 reference

### DIFF
--- a/_docs/releasing_version.md
+++ b/_docs/releasing_version.md
@@ -6,10 +6,13 @@ summary: How to perform a complete version release, including modulesync and pub
 ---
 
 Creating a release is a two step process:
+
 1. Prepare the release — setup everything so that peer review can happen and when everything is ready…
 2. Do the actual release
 
 ## Preparing a release
+
+This has to be done on a __*private*__ fork.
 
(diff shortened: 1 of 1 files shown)


_docs/releasing_version.md
```diff
@@ -6,22 +6,25 @@ summary: How to perform a complete version release, including modulesync and pub
 ---
 
 Creating a release is a two step process:
+
 1. Prepare the release — setup everything so that peer review can happen and when everything is ready…
 2. Do the actual release
 
 ## Preparing a release
 
+This has to be done on a __*private*__ fork.
```
bastelfreak, line 15: maybe personal instead of private? Because GitHub also has the option to make a repository private/public, this sentence could confuse people. And it's not technical required to do it on a fork, but we recommend it:
```suggestion
This should be done on a __*personal*__ fork.
```

rwaffen: yes its not techn. needed, but everybody does it like this and most people are not allowed to push branches to our repos? are they?

PR #333, opened 2023-12-13
Mention Puppet/Perforce sponsoring

bastelfreak:


_sponsors/puppet.md
```diff
@@ -0,0 +1,8 @@
+---
+name: Puppet by Perforce
+logo: sponsor-hetzner.png
```
vchepkov, line 3: Shouldn't it be sponsor-puppet?

bastelfreak: ah copy and paste :D updated it

vchepkov (approved):

PR #342, opened 2024-02-23
Update collaborators/repo counter

bastelfreak:


zilchms (approved):
Isnt that something we could automate somehow? ^^

bastelfreak:
We had (or still have?) javascript in place that pulls it live from GitHub. But now that only works if you'red logged into GitHub :(

It would be nice to have some automation to update the site.

zilchms:
building that as a workflow / periodic bot here? might be an idea

bastelfreak:
I'm happy to review your PR!

PR #345, opened 2024-03-05
Update software_support_lifecycle.md with new OS list

yakatz:
There should probably be some way to automate this, but in the mean time, this is an updated list of OS dates (except LEAP). The EndOfLife.software website also has not been updated, but EndOfLife.date has.


yakatz:
Based on the Git log, this table has not been updated since it was originally created: https://github.com/voxpupuli/voxpupuli.github.io/commit/e4b05abb7ccd15a564d22edd5c379477da3418e7

ekohl (commented):
I would like an enhancement to [puppet_metadata](https://github.com/voxpupuli/puppet_metadata) that uses [PuppetMetadata](https://github.com/voxpupuli/puppet_metadata/blob/master/lib/puppet_metadata/operatingsystem.rb) to split out this table. And one that if you use it with `metadata.json` that it tells you what to remove.

For completeness, I'd keep the older versions in. I'm not sure we've made sure all modules have been cleaned up and the historical note can be useful.

PR #350, opened 2024-03-12
Fix edit button consistency

smortex:
The button style is not consistent accross pages, so move the button in its own include file and rely on it where applicable.


bastelfreak:
would be nice if github could deploy it to a staging page.

PR #365, opened 2024-04-12
Add info to webpage how to get in touch

rwaffen:
is the page title okay? or should i use just "Contribute"?

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -25,6 +25,8 @@ collections:
     output: false
 
 navigation:
+  - title: How to contribute
+    url: /contributing/
   - title: Code of Conduct
     url: /coc/
   - title: Documentation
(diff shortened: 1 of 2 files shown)


bastelfreak:
I think such a page makes sense. Three ideas:
* Should we mention that we've more specific documentation at https://voxpupuli.org/docs/ ?
* Should we mention how people can contact us (like IRC/mailinglist/slack)?
* Mention that every project has a contributing guideline with more specific instructions?

smortex:
> is the page title okay?

A "long" title for the page looks fine to me.  Maybe if we add it as a menu item in https://github.com/voxpupuli/voxpupuli.github.io/blob/0c25777c12f2317b8040cbe467d772a0741efce7/_config.yml#L27 we can use only "Contribute".





> Should we mention how people can contact us (like IRC/mailinglist/slack)?

It is in the footer of all pages.  I feel like this is enough but I am probably biased.

bastelfreak (approved):

PR #386, opened 2024-10-01
Consider Debian LTS as supported

bastelfreak:
In the past we agreed on not supporting Debian LTS versions, only the standard versions (and I know that I was one of the people that pushed for not supported LTS back then). Debian 11 switched from standard to LTS on 2024-08-15. There are still requests from users to support Debian 11 and I think with the slow response from Puppet about new debian packages it makes sense to reconsider it.

Please respond with :+1: if you agree or with :-1: if you think we shouldn't support the LTS versions. When we've some feedback we can merge this PR or close it.

### DIFF
--- a/_posts/2024-10-01-debian-11.md
+++ b/_posts/2024-10-01-debian-11.md
@@ -1,0 +1,20 @@
+---
+layout: post
+title: Debian LTS support in our modules
+date: 2024-10-01
+github_username: bastelfreak
+category: changelog
+---
+
+In the past, Vox Pupuli decided against supporting [Debian LTS timeframes](https://wiki.debian.org/LTS).
+This was mostly because our CI resources were quite limited back in the days of Travis CI.
+Over the time we got multiple requests from individuals and companies to also support Debian LTS, some of them even willing to sponsor us.
+Some of our modules rely on a puppetserver or bolt, for example [puppet-redis][redis].
+We're now in a situation where Perforce fails to deliver packages in time (see [1][1], [2][2], [3][3]).
+This now brings us to a situation where normal Debian 11 support is gone, but we cannot add Debian 12 support for some modules because Perforce doesn't provide required packages.
+Therefore we consider Debian LTS releases as supported.
+
+[redis]: https://github.com/voxpupuli/puppet-redis/blob/master/spec/acceptance/redis_cli_task_spec.rb
+[1]: https://github.com/puppetlabs/bolt/issues/3280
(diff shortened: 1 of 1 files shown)


ekohl:
As I've said in private, I view Vox Pupuli as a collective of people who make sure they can do what they need. we're limited on the time we have and the reason I supported active removal of older code was to reduce maintenance burden. All in order to save time. If you actively want to support Debian 11 and put in the effort to make that happen, I'm all in favor of it.

The same goes for other cases. As long as it's reasonable (so don't ask for Puppet 3 support) and doesn't put a big burden on other maintainers then I'm open to it.

jay7x:
I'd announce this a best effort exception for now, so nobody would expect the same in the future just in case

bastelfreak:
Indeed, all the operating systems we "support" are best effort.

_posts/2024-10-01-debian-11.md
```diff
@@ -0,0 +1,20 @@
...
+layout: post
+title: Debian LTS support in our modules
+date: 2024-10-01
+github_username: bastelfreak
+category: changelog
+---
+
+In the past, Vox Pupuli decided against supporting [Debian LTS timeframes](https://wiki.debian.org/LTS).
+This was mostly because our CI resources were quite limited back in the days of Travis CI.
+Over the time we got multiple requests from individuals and companys to also support Debian LTS, some one them even willing to sponsor us.
```
kenyon, line 11:
```suggestion
Over the time we got multiple requests from individuals and companies to also support Debian LTS, some of them even willing to sponsor us.
```

bastelfreak: thanks, that happens when I had too much or not enough caffeine :D

ekohl (approved):

b4ldr (approved):

b4ldr:
I support this move but want to ask what this means for the puppet versions we support.  For instance the current LTS version, buster, is using puppet 5.5 support for which was removed from the voxpupuli modules long ago.  Or is the stance that we only support debian LTS with the perforce provided packages?

bastelfreak:
@b4ldr thanks for raising this. My understanding is that for distributions where Perforce offers AIO packages, we only support those. But I realize that we probably never explicitly wrote that down. I will try to come up with something in another PR.

At the moment, our beaker jobs all rely on AIO packages, where available. For Ubuntu 24.04 and Debian 12 we started with the distro packages bevause Perforce was a bit slow with packages. UT as soon as they were available, we switched to them.

jay7x (approved):

PR #404, opened 2024-12-27
use the unstyled menu list on mobile

binford2k:
The dropdown looks great on full size, but when we switch to the mobile
view, the unstyled list actually looks better. This breakpoint matches
the `navbar-expand-lg` parent class, so if that's changed, then this
should change to match. See https://getbootstrap.com/docs/5.0/layout/breakpoints/#available-breakpoints

Note that this will be easier to review if you set to ignore whitespace. All this does is add a breakpoint and indent some code to match.

### DIFF
--- a/static/css/site.css
+++ b/static/css/site.css
@@ -51,37 +51,42 @@ ul.docs-index li:hover {
   padding-top: 30px;
 }
 
-.navbar .submenu {
-  position: relative;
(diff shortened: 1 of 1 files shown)


binford2k:
Screenshot 2024-12-27 at 11 00 33 AM

bastelfreak (approved):

PR #411, opened 2025-01-14
Fix when we show next month meeting

smortex:
If the current month meeting has started 1h ago or more (and not is
going to start in less that one hour), it is over, and we wnat to move
to the next month.


smortex:
I currently see:

> Next monthly sync: 02/11/2025 (2025-02-11T05:30:00-10:00)

With the this I have the expected:

> Next monthly sync: Today at 5:30 AM (2025-01-14T05:30:00-10:00)

```
romain@zappy ~ % date
Tue Jan 14 05:04:47 -10 2025
```

genebean (approved):

PR #414, opened 2025-01-15
Update the install instructions for the new packages

binford2k:
The `openvox-server` and `openvox-db` packages haven't been published
yet, but they should be very shortly.

### DIFF
--- a/openvox/install.md
+++ b/openvox/install.md
@@ -51,11 +51,11 @@ Then install the packages you want.
 * Debian family
   * `apt install openvox-agent`
   * `apt install openvox-server`
-  * `apt install openvox-db`
+  * `apt install openvoxdb`
 * RedHat family
   * `yum install openvox-agent`
   * `yum install openvox-server`
-  * `yum install openvox-db`
+  * `yum install openvoxdb`
 
 If you have backed up config files, then restore them now.
 
--- a/openvox/install.md
+++ b/openvox/install.md
@@ -5,16 +5,40 @@ subsection: openvox
 ---
 
 OpenVox package downloads are currently sponsored by [Overlook InfraTech](https://overlookinfratech.com).
-Following Perforce's suggestions, these are straight rebuilds of the original packages with no branding or name changes.
-Be aware that this will change as soon as the OpenVox pipelines are running.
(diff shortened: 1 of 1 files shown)


openvox/install.md
```diff
@@ -5,16 +5,40 @@ subsection: openvox
...
+We do not yet have a fully robust test pipeline, although that is our next priority.
+
+
+## Uninstalling Puppet
+
+We encourage you to try out OpenVox on a fresh test system, the way you would for any major system package.
+If you'd rather try it on an existing system or develop a migration process, then you will first have to uninstall Puppet.
+
+* If you're migrating from Puppet Enterprise you can use the `puppet-enterprise-uninstaller` script on each node as described in [their docs](https://www.puppet.com/docs/pe/latest/uninstalling.html).
+* If you're using the all-in-one packages such as `puppet-agent` or `puppetserver` from the `[apt|yum].puppet.com` repos, simply remove these packages.
```
ekohl, line 22: Do we have rpm/deb metadata that says it conflicts? Possibly even provides and automatically replaces the Perforce packages

binford2k: @nmburgan is working on the conflicts/replaces metadata. I *think* that will only be sufficient for the simple "replace the AIO package" case though. I don't think that removes unused dependencies. (but I could totally be wrong)

openvox/install.md
```diff
@@ -5,16 +5,40 @@ subsection: openvox
 ---
 
 OpenVox package downloads are currently sponsored by [Overlook InfraTech](https://overlookinfratech.com).
-Following Perforce's suggestions, these are straight rebuilds of the original packages with no branding or name changes.
-Be aware that this will change as soon as the OpenVox pipelines are running.
+As of release 8.11, OpenVox is functionally equivalent to Puppet; the command names are the same, the configuration file paths are the same, etc.
+The major differences are in help text output, man pages, and so on.
```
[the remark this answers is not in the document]
binford2k: close enough. We'll probably be replacing text throughout the whole year, lol

openvox/install.md
binford2k: @ekohl are you ok merging as-is? We can always come back and simplify if the metadata does indeed do the heavy lifting for us.

binford2k: merging as-is now, so we have install instructions that at least use the right names. We'll come back to it again after next release.

PR #430, opened 2025-02-25
Example for Yum

zoojar:
added an example for enabling repo and installing on Yum


jstraw (changes requested):

bastelfreak:
I think the changes here are obsolete now. Please rebase them if you think it's still required.

PR #432, opened 2025-02-26
Use sticky PR comment only for branches from origin

bastelfreak:
the action only works when the base branch is in the Vox Pupuli org.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,7 +31,9 @@ jobs:
           overwrite: true
           compression-level: 9
           if-no-files-found: error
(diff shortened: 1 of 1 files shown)


ekohl:
You also tried https://github.com/marocchino/sticky-pull-request-comment?tab=readme-ov-file#error-resource-not-accessible-by-integration I suppose.

jstraw:
it didn't seem to work when I was testing it, but I also didn't have access to change the repo configuration, so didn't attempt to push further on it.

jstraw (approved):

PR #439, opened 2025-03-05
pql_queries: add "get nodes with specific array element value in a fact"

kenyon:
Could not find an example of doing this. I could not find a way to filter arrays of structured facts with PQL, so had to use `jq`.


bastelfreak:
@kenyon thanks for the PR! Can you please add a gpg or ssh signature to the commit?

PR #446, opened 2025-04-09
Add ATIX to support page

bastelfreak:


_support/atix.md
```diff
@@ -0,0 +1,7 @@
+---
+logo: sponsor-atix.png
```
ekohl, line 2: Is ATIX sponsoring? I though we only wanted to list them as a party that can provide support. IMHO we should provide a clear separation between sponsors and non-sponsors.

bastelfreak: Yes. They will be listed at the support page, not the sponsoring page (this is at /_support/, not _sponsors). But the Makefile that generates the png files currently requires all logos to be prefixed with `sponsor-`.

ekohl: Perhaps introduce a new `logo-%.png` rule?

bastelfreak: updated it

ekohl (approved):
From a structure point of view :+1: but I'll leave it open for ATIX to review the actual text.

bastelfreak:
They send me the text via email, so I assume it's fine.

PR #454, opened 2025-04-24
Add openvox pp yum install

yakatz:
This is from Slack, but I have not personally tested

### DIFF
--- a/openvox/install.md
+++ b/openvox/install.md
@@ -79,6 +79,21 @@ apt::source { 'openvox8-release':
 }
 ```
 
+```puppet
+include yum
+$release=8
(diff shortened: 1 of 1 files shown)


openvox/install.md
```diff
@@ -79,6 +79,21 @@ apt::source { 'openvox8-release':
 }
 ```
 
+```puppet
+include yum
+$release=8
+$os_name = $facts['os']['name'] ? {
+  /(AlmaLinux)|(Rocky)|(RedHat)|(Centos)/ => "el",
```
bastelfreak, line 86: Maybe use default here instead, since 99% of the Redhat Derivates out there are el

PR #463, opened 2025-05-08
Simplify the install page

binford2k:
The install page was starting to look a lot more complex than it had to
be. This simply focuses on the "easy path" and moves everything else down
to an "advanced" section.


openvox/install.md
```diff
@@ -57,9 +40,34 @@ Then install the packages you want.
 
 If you have backed up config files, then restore them now.
 
+### 🎉 That's it! You're done!
```
bastelfreak, line 43: should we add a manual ToC to the document? Or there's probably an option to render it automatically? :thinking:

binford2k: There are plugins to generate them, but I didn't feel like taking the time to find and configure one. We can do that another time and revamp the docs sections too.

Screenshot 2025-05-08 at 12 18 22 PM

PR #476, opened 2025-05-22
Rough draft of new voxconf page

binford2k:
This looks all right at a desktop resolution, but the info overlay
floats around quite a bit as resolution changes, so it needs a bit of
work still.

I'd love it if someone with a bit more time picked this up and finished it!

Screenshot 2025-05-22 at 3 09 35 PM

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -67,6 +67,9 @@ navigation:
       - title: Other Channels
         url: /connect/
         icon: fas fa-comments
+      - title: VoxConf
+        url: /voxconf/
+        icon: fas fa-person-chalkboard
       - title: Code of Conduct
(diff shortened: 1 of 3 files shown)


binford2k:
Most of the site is bootstrap. I didn’t find a helper class for the overlay that didn’t require writing the whole thing in raw html though :-/

sebastianrakel:
> Most of the site is bootstrap. I didn’t find a helper class for the overlay that didn’t require writing the whole thing in raw html though :-/
> 
> I would be ecstatic if you found the right classes though 😝

Yeah... i used html now to center at bottom :D tell me if you like it

PR #479, opened 2025-05-26
Improve layout on mobile

defnull:


bastelfreak:
@defnull can you please your commit with an ssh or gpg key?

PR #482, opened 2025-05-27
VoxConf: Add registration URL

bastelfreak:


voxconf.md
```diff
@@ -28,7 +28,7 @@ title: VoxConf '25
     </tr>
     <tr>
       <th scope="row">Register:</th>
-      <td><p class="fst-italic">coming soon!<br/>👇Scroll down for more information⬇️</p></td>
+      <td><a href="https://tickets.netways.de/NES/foremanbirthday/">click here!</a></td>
```
binford2k, line 31: my original intent here was to have both the link and the scroll down message

PR #517, opened 2025-07-22
Add new section for former sponsors.

binford2k:
This just creates a new section with smaller cards for former sponsors.
To move a card there, just add `status: former` to the front matter.

### DIFF
--- a/static/css/site.css
+++ b/static/css/site.css
@@ -118,11 +118,6 @@ section.content {
   margin: 1em 0;
 }
 
-.former .card {
-  zoom: 0.65;
-  width: 20em;
-}
-
 .callout {
   width: 45%;
   float: right;


static/css/site.css
```diff
@@ -118,6 +118,11 @@ section.content {
   margin: 1em 0;
 }
 
+.former .card {
+  zoom: 0.65;
+  width: 20em;
+}
+
```
smortex, line 125: I would avoid tweaking CSS if possible, that would mean not adding this…

sponsoring/index.md
```diff
@@ -28,7 +28,21 @@ At the moment we've multiple sponsors via GitHub. In addition to that, multiple
...
+</div>
+
+---
+
+## Former Sponsors
+
+<div class="row former">
+  {% for sponsor in site.sponsors -%}
+    {% if sponsor.status == "former" -%}
+    {% include sponsor-card.html width=4 vendor=sponsor %}
```
smortex, line 45: …and use less columns here, e.g.:

```suggestion
    {% include sponsor-card.html width=3 vendor=sponsor %}
```

(untested, IIRC there are 12 columns, so width=4 means 3 logos per line, while width=3 means 4… we can even go to width=2)

sponsoring/index.md
```diff
@@ -28,7 +28,21 @@ At the moment we've multiple sponsors via GitHub. In addition to that, multiple
 companies sponsor our cloud/CI resources.
```
smortex, line 29: If we have a "Former Sponsors" bellow, a "Current Sponsors" here would be great no?

```suggestion

## Current Sponsors

```

smortex (commented):
Not sure if we should keep former sponsors at all, but if so I added minor in-line comments.

static/css/site.css
binford2k: heh, clearly wasn't thinking much here 😀

PR #518, opened 2025-07-22
Add social media accounts

binford2k:
Also revamped the page just a little bit, it was starting to look
crowded.


smortex (approved):
That's a lot of icons :smile: 

LGTM

binford2k:
@smortex if we decide the icons look silly, I've no attachment. We can always rip them back out. Just trying to break up a page that shouldn't feel so dense 😀

PR #520, opened 2025-07-25
Update VoxConf '25 schedule with speaker slides; add recordings section

rwaffen:


voxconf/2025.md
```diff
@@ -68,15 +68,27 @@ Connect with us on our Slack in #voxconf2025 and/or on IRC in #voxpupuli-voxconf
 
 {: .table .table-striped .table-hover .w-75}
```
bastelfreak, line 69: with the 75char limit we now have a line break. Is that intentional or should the table become wider?

rwaffen: oh, not intentional. will check this again.

for next voxconf: enforce shorter titles 😆

rwaffen: i would say we drop the 75%-rule because the tables is so wide already and if we add also the recording links to each talk it will be even longer.

PR #527, opened 2025-08-01
Update quickstart.md - use https and point directly to the right docs

SvenVD-be:
Make it easier for user to find the correct documentation


bastelfreak (approved):

smortex:
Given that can go in any direction, is it not better to point to an archived version of a compatible version of puppet, e.g. https://web.archive.org/web/20241108102350/https://www.puppet.com/docs/puppet/8/puppet_index

PR #534, opened 2025-08-18
Added information on installing Foreman w/ OpenVox

binford2k:
The big trick is that you have to start by installing OpenVox and then
installing the oauth gem. Then the installer just works.

This page is getting a little unwieldy again. We'll need another cleanup
pass on it soon.

### DIFF
--- a/openvox/install.md
+++ b/openvox/install.md
@@ -97,11 +97,15 @@ And finally, use [Foreman's Quickstart Guide](https://theforeman.org/manuals/3.1
 
 ### Porting a Foreman infra to OpenVox
 
+> 🚨 Tip: Resetting the Foreman answers file will remove any customization you're currently using, so you'll need to specify all your desired options again.
+> We suggest backing up the file and then diffing it against the new version to see what changed.
+{: class="alert alert-warning callout" }
+
 If you're already running Foreman and you want to switch it to be backed by OpenVox then you have a little bit of housekeeping to do first.
 First, use your system package manager to remove the `puppet-agent-oauth` package.
 If you use the `foreman-installer` to manage and upgrade your setup, then ensure that you restore the cached answers file to the default values, or it will continue to use old paths.
 
-<details class="card" >
+<details class="card" style="clear: both;">
 <summary class="card-header">Debian Family</summary>
 <pre><code># rm /etc/foreman-installer/scenarios.d/foreman-answers.yaml
 # apt install --reinstall -o Dpkg::Options::="--force-confmiss" foreman-installer</code></pre>
@@ -118,7 +122,11 @@ If you use the `foreman-installer` to manage and upgrade your setup, then ensure
 </details>
 
 If you're porting from system installed Puppet, then you may have to purge package configuration or remove legacy directory structures.
-For example, you might `rm -rf /etc/puppet`
+For example, you might `rm -rf /etc/puppet`, making sure to back up or move critical configuration files to new locations.
+Make sure you ensure that files like `puppet.conf` don't reference legacy locations.
+Note that in most cases, paths don't actually need to be specified because the defaults are correct.
+
+Double check that you have no `puppet`, `puppet-agent`, or any other Puppet related packages installed and that at least `openvox-agent` is installed as per the instructions at the top of the page.
 
 Ensure that the `aio_agent_version` fact returns the appropriate version before continuing.
 Then follow the [installation instructions](#installing-foreman-with-openvox) above to update your infrastructure.


openvox/install.md
```diff
@@ -50,28 +53,62 @@ Then install the packages you want.
...
+puppet::client_package: openvox-agent
+puppet::server_package: openvox-server
+puppetdb::puppetdb_package: openvoxdb
+puppetdb::master::config::terminus_package: openvoxdb-termini
+foreman::providers::oauth: false
+```
+Then manually install the `oauth` gem into OpenVox's gempath.
+
+```
+# /opt/puppetlabs/puppet/bin/gem install oauth
```
ekohl, line 87: This isn't needed anymore now that `openvox-agent` provides `puppet-agent`
```suggestion
```

openvox/install.md
```diff
@@ -50,28 +53,62 @@ Then install the packages you want.
...
+If you're already running Foreman and you want to switch it to be backed by OpenVox then you have a little bit of housekeeping to do first.
+First, use your system package manager to remove the `puppet-agent-oauth` package.
+If you use the `foreman-installer` to manage and upgrade your setup, then ensure that you remove the cached scenario file, or it will continue to use old package names.
+
+```
+# rm /etc/foreman-installer/scenarios.d/last_scenario.yaml
+```
+
+If you're porting from system installed Puppet, then you may have to purge package configuration or remove legacy directory structures.
+Ensure that the `aio_agent_version` fact returns the appropriate version before continuing.
```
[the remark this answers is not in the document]
ekohl: It really doesn't matter what it returns, just that it's a string to indicate it's an AIO build: https://github.com/theforeman/puppet-foreman/blob/4ac87a6d84e860f1bcdd402a67141b5609c9190b/manifests/params.pp#L37

openvox/install.md
```diff
@@ -50,28 +53,62 @@ Then install the packages you want.
...
+#### Using the `theforeman-foreman` module
+
+1. Ensure that OpenVox is installed, then follow the instructions in the [Managing OpenVox with OpenVox](#managing-openvox-with-openvox) sections to configure Hiera properly.
+1. Add `theforeman-foreman` to your `Puppetfile` and deploy it.
+1. Use the [documentation](https://github.com/theforeman/puppet-foreman) to classify your server node to your requirements and then run the OpenVox agent to configure everything.
+
+### Porting a Foreman infra to OpenVox
+
+If you're already running Foreman and you want to switch it to be backed by OpenVox then you have a little bit of housekeeping to do first.
+First, use your system package manager to remove the `puppet-agent-oauth` package.
```
ekohl, line 101: Not needed anymore.
```suggestion
```

openvox/install.md
```diff
@@ -50,28 +53,62 @@ Then install the packages you want.
...
-You do not need to remove the `[apt|yum].puppet.com` repositories although the only thing you'll be able to use them for going forward is installing historical Puppet releases.
+#### Using the `foreman-installer`
+
+Edit `/etc/foreman-installer/custom-hiera.yaml` to include these Hiera settings.
+If you intend to continue managing your Foreman installation with puppet code, then follow the instructions in the [Managing OpenVox with OpenVox](#managing-openvox-with-openvox) sections to persist these settings into your standard Hiera data files.
+
+``` yaml
+---
+puppet::client_package: openvox-agent
+puppet::server_package: openvox-server
```
ekohl, line 79: These are answers and shouldn't be overridden via Hiera. The real way is to provide them options:

```
foreman-installer --puppet-client-package openvox-agent --puppet-server-package openvox-server
```

This also should solve your "caching" issue with the answers.

openvox/install.md
```diff
@@ -50,28 +53,62 @@ Then install the packages you want.
...
+
+### Porting a Foreman infra to OpenVox
+
+If you're already running Foreman and you want to switch it to be backed by OpenVox then you have a little bit of housekeeping to do first.
+First, use your system package manager to remove the `puppet-agent-oauth` package.
+If you use the `foreman-installer` to manage and upgrade your setup, then ensure that you remove the cached scenario file, or it will continue to use old package names.
+
+```
+# rm /etc/foreman-installer/scenarios.d/last_scenario.yaml
+```
```
ekohl, line 106: This is never a good idea. It's a symlink pointing to the real scenario. This deselects the scenario, but that really isn't of any use. Especially if you only have a single scenario available.

Nor is it needed. The scenario file is describing the scenario to be used. The answers file (which you're not using) is 
```suggestion
```

ekohl (changes requested):
This is a quick review of things that jumped out, but it's late and I may be missing something.

openvox/install.md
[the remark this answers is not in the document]
binford2k: Yes. but it will either return something (accurate) or nothing (inaccurate) and this sounded a bit more trustworthy than "returns a value". 

Also, if it genuinely _does_ return something that's a wrong value, then it indicates something else is broken and should be investigated.

openvox/install.md
binford2k: 8.18.1 is only a few months old. I don't think we can drop docs for it yet. I'll make a note about this not being needed on current though.

openvox/install.md
binford2k: Hrm. This gets complex when describing how to do something as simple as including the `puppetdb` plugin. As far as I can tell, there's no CLI options for 

``` yaml
puppetdb::puppetdb_package: openvoxdb
puppetdb::master::config::terminus_package: openvoxdb-termini
```

openvox/install.md
binford2k: I'm not convinced this it's enough to specify the CLI parameters. I can validate later, but what happens now is that someone on Debian does

1. apt install foreman-installer (which pulls in Debian packaged `puppet-agent`, non-AIO)
2. configures hiera data (which I believe is roughly equivalent to passing CLI args)
3. runs the installer
4. everything breaks because `/etc/puppet/<stuff>` exists even after the `openvox-agent` package replaces the system `puppet-agent` package and foreman-installer continues to use the `/etc/puppet` directory.

Even after deleting the `/etc/puppet` directory it still breaks, even with the proper package names. That's what we're trying to address here.

openvox/install.md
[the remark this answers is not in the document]
ekohl: Fair enough.

openvox/install.md
binford2k: Yes, if you install the proper package to begin with, then it works great. But we can't require perfection from our users! The only way to get from system `puppet-agent` -> AIO `openvox-agent` is to start with a fresh answers file or to edit some 500 lines of data and correct each and every legacy path.

openvox/install.md
```diff
@@ -50,28 +53,75 @@ Then install the packages you want.
...
+
+### Porting a Foreman infra to OpenVox
+
+If you're already running Foreman and you want to switch it to be backed by OpenVox then you have a little bit of housekeeping to do first.
+First, use your system package manager to remove the `puppet-agent-oauth` package.
+If you use the `foreman-installer` to manage and upgrade your setup, then ensure that you restore the cached answers file to the default values, or it will continue to use old paths.
+
+<details class="card" >
+<summary class="card-header">Debian Family</summary>
+<pre><code># rm /etc/foreman-installer/scenarios.d/foreman-answers.yaml
```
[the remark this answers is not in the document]
binford2k: tbh, if they're reading this part of the docs, they're _already in_ a broken state.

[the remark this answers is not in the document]
binford2k: Also, even with correct metadata, that doesn't fix the fact that foreman-installer still tries to use the incorrect paths in the cached answers file!

ekohl:
Today we made progress on the Debian side:
* https://github.com/OpenVoxProject/openvox-agent/pull/120 & https://github.com/OpenVoxProject/vanagon/pull/51 add the version number for Provides on Debian. That should help it to properly replace the existing package
* @tuxmea found we need to drop the `Conflicts` on Debian for `foreman-installer` to accept it as a proper replacement.

openvox/install.md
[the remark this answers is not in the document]
binford2k: @ekohl Came back and poked a little bit this week. It seems now that this is sufficient on redhat flavored machines that don't have the puppet repos enabled.

1. `dnf install https://yum.voxpupuli.org/openvox8-release-el-9.noarch.rpm`
2. `dnf install https://yum.theforeman.org/releases/3.16/el9/x86_64/foreman-release.rpm`
3. `dnf install foreman-installer`
4. `foreman-installer --puppet-client-package openvox-agent --puppet-server-package openvox-server`

But it still doesn't seem to correct _derived values_ that get cached in the answers file if you mess up the first time around. What's Foreman's approach to fixing those?

My preference is still to reset the answers file and recreate it. I don't see the possibility of breaking a system being a problem because they've just installed and can just ⬆️  a few times and re-run the installer.

openvox/install.md
bastelfreak: @binford2k do you still see issues? foreman + openvox is working fine for me without manually installing the gem

PR #536, opened 2025-08-25
Document voxpupuli-github-notifications channel

bastelfreak:


h-haaks (approved):
Should all the pipes be alligned?

kenyon (approved):
Aligning the table formatting just causes unnecessary diffs IMO, similar to aligning the equals signs in puppet class parameter lists ;)

PR #540, opened 2025-08-31
support lifecycle: remove OS support table

kenyon:
This table was not being updated, so we should just remove it.

### DIFF
--- a/_docs/software_support_lifecycle.md
+++ b/_docs/software_support_lifecycle.md
@@ -31,6 +31,9 @@ An Operating System is no longer considered supported once it reaches an EOL dat
 
 *maintainers include, but are not limited to, the community, company, or organization that maintains ownership over a specific piece of software or project.
 
+The [puppet_metadata](https://github.com/voxpupuli/puppet_metadata/) gem is the authoritative source for EOL data used in the Vox Pupuli testing framework.
+Tests will only run for operating systems that are still supported according to the puppet_metadata gem.
+
 Sources for end-of-life dates:
 
 * OpenSUSE - [OpenSUSE Documentation](https://en.opensuse.org/Lifetime)


_docs/software_support_lifecycle.md
```diff
@@ -31,32 +31,6 @@ An Operating System is no longer considered supported once it reaches an EOL dat
...
-|     Fedora 32      |   2020/04/21   | 2021/05/31 (approx) | 2021/02/28 (approx) |   2021/05/31   |   2021/05/31    |
-| OpenSUSE Leap 15.1 |   2019/05/22   |     2021/01/31      |     2020/12/31      |   2021/01/31   |   2021/01/31    |
-| OpenSUSE Leap 15.2 |   2020/07/02   |     2021/12/31      |     2020/09/30      |   2021/12/31   |   2021/12/31    |
-
-NOTE: *\*All Dates subject to change based on upstream project*
-
-Sources:
-
-* OpenSUSE - [OpenSUSE Documentation](https://en.opensuse.org/Lifetime)
-* All others - [EndOfLife.software](https://endoflife.software/operating-systems)
```
ekohl, line 58: Probably good to still keep these links in for the OSes. We don't have to include a table, but rather a way to figure it out. https://github.com/voxpupuli/puppet_metadata/blob/master/lib/puppet_metadata/operatingsystem.rb also contains the data.

ekohl (changes requested):

_docs/software_support_lifecycle.md
kenyon: I considered leaving those links. Updated this PR to keep them.

bastelfreak:
@kenyon should we mention that puppet_metadata` is the authoritative tool? That's what we use in CI, we only spawn Beaker jobs for distributions that's aren't marked EoL by puppet _metadata.

PR #561, opened 2025-10-19
Add community changelog

bastelfreak:
This is a collection of all Git tags and GitHub releases from various organisations within the Puppet/OpenVox ecosystem. All releases are sorted by date.


bastelfreak:
I'm working on a script to get all the releases from the ecosystem. I think this helps to visualize how active the community is. I'm still working on the markdown layout and the script itself. As soon as it works properly, I will add it as well to the PR. I already raised the PR so people can get a first impression and provide feedback. I also want to add the puppetlabs org, but github API ratelimits are... a thing.

You can render this locally with `bundle exec jekyll serve -w --config _config.yml,_config-dev.yml`

changelog/index.md
ekohl: Do we really want to store such a huge file in git? Perhaps create a `changelog.voxpupuli.org` and render it there as its own website? Then we can also apply some more aggressive automation to update the site.

PR #589, opened 2025-12-02
feat: update obsolete foreman/openvox integration settings

roock:
The openvox-agent package now has the provide puppet-agent flat on Debian and Redhat, so we can remove the workaround for foreman about puppet-agentt-oauth package not being installable with openvox.

``` 
# rpm -q --provides openvox-agent
openvox-agent = 8.23.1-1.el9
openvox-agent(x86-64) = 8.23.1-1.el9
puppet-agent = 8

# apt-cache show openvox-agent |grep puppet-agent  
Conflicts: puppet-agent 
Breaks: puppet-agent (<< 9) 
Replaces: puppet-agent (<< 9) 
Provides: puppet-agent (= 8)
```

### DIFF
--- a/openvox/install.md
+++ b/openvox/install.md
@@ -134,23 +134,20 @@ puppetdb::master::config::terminus_package: openvoxdb-termini
 Note that you will need to have the OpenVox repositories available (using one of the methods above) for this to work properly.
 If the OpenVox repositories are available, this will cause Puppet to be removed and OpenVox to be installed.
 
-
 #### Foreman integration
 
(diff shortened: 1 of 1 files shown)


yakatz:
Would it be better to leave the heading and say underneath it that Foreman will work properly with no workaround required anymore?

Otherwise, people will remember that there was once such an issue and will spend time trying to find it again.

roock:
@bastelfreak  @yakatz  I've updated the PR with your suggestions

PR #671, opened 2026-05-12
Add AI policy to transfer checklist

binford2k:


kenyon (approved):

_docs/migrate_module.md
```diff
@@ -57,6 +59,20 @@ If the owner has responded and is not interested in migrating their module to VP
 To start the process, document your request and efforts in a brief email to the [mailing list](https://groups.io/g/voxpupuli/).
 If the module is accepted, VP will work with you to determine the proper fork/migration steps needed in addition to the checklist above.
 
+## Adding AI trailers
+
+If you're contributing a project with commits that need to have the `Assisted-By:` AI attribution trailers, you can do that with an interactive rebase.
```
bastelfreak, line 64: should people do this to their main branch before migrating a git repo? That would rewrite their entire history and break existing checkouts.

smortex: I agree, that is not great.

Having old commits not complying with our standards is not better, but maybe preferable?

Also, we can recommend to create a branch that is basically the state of main before rewriting history with a name that help figure out why such a branch exist (e.g. "main-before-complying-with-voxpupuli-ai-policy").  The hope is that when somebody who has been working of the repo tries to git pull and get a failure because history was re-written, they now see their commits on top of this weird branch name.  And how they know that they car git reset --hard main; then git cherry-pick each commit; and finally git push -f is still to be determined 😨…

bastelfreak: I really don't think this is worth the effort. I'm not aware of other projects had rewrite the entire history.

ekohl (changes requested):
Agreed we shouldn't rewrite history on transfer. Otherwise we might as well fork the project.

binford2k:
Do we just let it go then, or do we request something like `HISTORY.md`?

